PR #4, opened 2018-01-22
Added project files for oXygen and XMLspy

syversenkr:
As commented by @nick-knowles and @Aurige, we should keep the XML tool project files in the main repository


skinkie:
@syversenkr please rename the files so they do not contain version information.

Have you checked the files in the repository 'work'? Hence no files are missing that are referenced form the XMLSpy project?

Aurige:
I have checked the XMLSpy one, it is Ok

syversenkr:
Rechecked, was issues with some WSDL paths, will recommit once fixed

syversenkr:
Have renamed the project files without version info, as well as cleaning up some invalid paths in the oXygen version
@Aurige please verify that renaming the XMLspy project file accordingly does not break any of its references

PR #5, opened 2018-02-14
Remove unused netex xsd files in v1.0 which also now exists in v1.1

csolem:
I have found some duplicated files. New files with v1.1 was added, but v1.0 versions of the files were not removed.

* acsb_accessibility
* responsibilitySet_version
* country_support
* submode_version
* distanceMatrixElement_version

The version pattern in file names should not be necessary when using git.

I found these duplicated files by running a bash one-liner (which also checks for references):
```
find . -iname "*xsd" | while read file; do if [[ $file = *"v1.1.xsd" ]]; then oldVersion=$(echo $file | sed -e 's/v1\.1/v1\.0/g'); if [[ -f $oldVersion ]]; then echo $oldVersion; folderLess=$(echo $oldVersion | sed 's@.*/@@'); grep -R "$folderLess" .; fi;  fi; done
```


syversenkr:
As a second precaution, I also diffed the files and checked manually checked the references, and as far as I could tell all the affected (i.e. removed) files are outdated and not in use.

skinkie:
I would be happy to merge this. But I rather have it solved like I did before. Remove the version from the filename. And properly update the file while maintaining history.

Aurige:
Yes, we still have a few old deprecated files, they are not in the XML Spy project from which we were previously managing the XSD, so they were a bit hidden. It's of course better to remove them when you discover one.

skinkie:
Removing kills the changes, hence don't do that. Rename the old file to the new version number, overwrite it with the new contents. And ofcourse apply a mega change to remove the suffix of all files.

syversenkr:
I vote for both, add this change now. Should not have any negative effects... no effect whatsoever, actually.
Then we (should probably define who "we" beforehand) can add the mega-fix of renaming the files altogether, as this also means changing A LOT of internal references (include/import).

skinkie:
@syversenkr are you sure it doesn't have no negative effects? Because we wouldn't have the changes from -v1.0 to -v1.1?

syversenkr:
@skinkie 
From reviewing the files (manually), my assumption is that v1.1 are copies of v1.0 with 2017 CRs and other changes/additions from the work group incorporated. Where there are v1.0 and v1.1 in parallell, we have found no case where the v1.0 is in use, i.e. in effect v1.0 is deprecated and replaced by its v1.1

@nick-knowles may elaborate further on this?

As neither file versions have been properly handled in a versioning systems before this repository, I believe the best way to move forward will be to

1. Delete the orphaned v1.0s (they have new, and better, siblings) - _This CR_
2. Ignore the fact that we miss changes from v1.0 to v.1.1 (diff would have to be done manually, as both files now exist; however we do not have changes tracked from v0.X to v1.0 on any of the files, so I think we can live with that)
3. Remove the versioning filename postfix and change all references (import/include) accordingly - _New PullRequest to come (@csolem will work on this in the foreseeable future)_ => History prevails in Git
4. Leave proper change management up to Git from that point on, by tagging releases etc.

skinkie:
@syversenkr I actually want to overwrite 1.0 with 1.1, then solve the references. So I'll try to do it today.

PR #6, opened 2018-02-20
Fix PriceRuleStepResultStructure's id attribute

csolem:
Change PriceRuleStepResultStructure's id attribute to be of type *ObjectIdType* instead of referring to "id". Because jaxb generation of the model causes generated netex xml to have an undesired namespace prefix.

@syversenkr can probably elaborate.


syversenkr:
PriceRuleStepResultStructure is _ducumented_ as having an ID attribute (latest revision NeTEx Part 3, figure 252), however only has REF implemented (not documented). Assumed correct implementation of this aspect reflected in the PullRequest. 

On another note, the implementation of  xsd:complexType name="PriceRuleStepResultStructure" does **not** match current documentation of PriceRuleStepResult (i.e. december revision of NeTEx Part 3) and one/both should be corrected.

csolem:
Fixed conflicts. Ready for merging. @skinkie

PR #7, opened 2018-02-20
Set max occurs on GroupOfStopPlaces in groupsOfStopPlacesInFrame

csolem:
There is an issue with having more than one GroupOfStopPlaces in a frame.
This pull request sets `maxOccurs="unbounded"` in complexType groupsOfStopPlacesInFrame_RelStructure for the sequence of element GroupOfStopPlaces.


syversenkr:
Error has previously been mentioned briefly. Feature needed for Norwegian National Stop Registry.

skinkie:
I was actually waiting for @nick-knowles big pull request...

csolem:
Ok. Why not merge this first? It would probably be an easy merge for @nick-knowles as this pull request contains a small change in the xsd.

skinkie:
Because @nick-knowles made changes before all your changes, but was not able to commit or do a pull requests. Hence we will cause a big problem for him, if he needs to merge your changes too. I'll mail him.

PR #15, opened 2018-04-12
fix parent of AbstractDiscoveryRequestStructure

muued:
AbstractDiscoveryRequestStructure extends AuthenticatedRequestStructure instead of RequestStructure to avoid duplicates fields of RequestorEndpointGroup


skinkie:
Should this also be fixed upstream?

muued:
Yes, that would be a good idea.

muued:
Seems to be fixed there, already

skinkie:
@nick-knowles can we merge this?

PR #24, opened 2018-08-09
Temporarily move invalid examples out to separate folder

csolem:
To make it possible to use travis for PR checks. Now it fails regardless.
Examples should be fixes in a separate PR (fyi: @nick-knowles)


skinkie:
Could you move them to the fail folder, like I did with the other examples?

PR #28, opened 2018-09-11
Format all xsd files using xmllint --format

csolem:
```
find examples/ -iname "*.xsd" | while read file; do echo "Formatting ${file}"; mv ${file} /tmp/backupfile; xmllint --format /tmp/backupfile > ${file}; done;
```


skinkie:
This is the command used before. where XMLLINT_IDENT is a tab. I do not want to loose history again on all files.

find . -name "*xsd" -type f | while read i; do XMLLINT_INDENT="	" xmllint --pretty 1 "$i" > "$i.pretty"; mv "$i.pretty" "$i"; done

csolem:
FYI @seime

PR #31, opened 2018-10-12
Remove ReadMe.txt file as README.md is probably the one to keep and use

csolem:


csolem:
Closed, as some lines in the txt might need to be moved over to the markdown file.

PR #35, opened 2019-01-03
Fix typo AccountingTIme

skinkie:

PR #36, opened 2019-01-04
Feature dutypart vehicleref

skinkie:
This is a WIP. But I would really would like to see some feedback :-)

### DIFF
--- a/xsd/netex_part_2/part2_vehicleService/netex_duty_support.xsd
+++ b/xsd/netex_part_2/part2_vehicleService/netex_duty_support.xsd
@@ -139,7 +139,7 @@ Rail transport, Roads and Road transport
 		</xsd:simpleContent>
 	</xsd:complexType>
 	<!-- ======================================================================= -->
-	<!-- === Driver ====================================================== -->
+	<!-- === DriverTrip ====================================================== -->
 	<xsd:simpleType name="DriverTripIdType">
 		<xsd:annotation>
 			<xsd:documentation>Type for identifier of a DRIVER TRIP.</xsd:documentation>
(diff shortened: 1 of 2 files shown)


Aurige:
Hello Stefan
I agree that these typos should be fixed, but this is also changing some tag names, and therefore may generate some incompatibilities with previous versions. So if we decide to do this, we need to have a very clear list of such changes to be validated with the group, and also anticipate some possible solutions (like substitution group) for those already using these tags.

skinkie:
@Aurige this is the pull you want to comment on.
https://github.com/NeTEx-CEN/NeTEx/pull/35

Aurige:
Yes, that's this issue, fixing **AccountingTIme** to **AccountingTime**, is something we should probably do, also it changes the tag name... with possible compatibility issue

skinkie:
You are commenting on #36 not #35.

ue71603 (approved):

skinkie:
@ue71603 this is certainly a next thing :-)

ue71603:
@skinkie  Ok, it drags now some master things with it, it seems (all changes to the examples).

duexw:
Some initial comments on duties and duty parts:

**DriverScheduleFrame:**
- It would be nice to have a reference to VehicleScheduleFrame

**Duty:**
- Should have number / external identifier (the duty number plays a strong role in existing systems)
- should have direct day type Reference like Block
- Reference to a Garage or to a DutyGroup which refers to a Garage
- either add start + end place and time or make dutyPart mandatory
- Qualifications would be good to have


**DutyPart:**
Driver Access / Return Duration: Included in Start/End Time or on top? 
What is the difference between this element and the driver trip times?

**dutyActivities:**
DriverTrips do not have explicit Start and End Time


**Driving Spell:** 
- Should also have Start and End Time (Reference to ReliefOpportunity is not enough)
- Spell/VehicleJourneyRef: A spell could cover several subsequent journeys


**General:**
AccountableElement would need some "Type of Account", e.g. an activity might lead to several times such as driving time, night time, inconvenience surcharge...
Standby / Reserve duty are two different types of duty, how would one model reserve duty?

Much better now, but some work to be done yet

skinkie:
Is addressed in #424

PR #38, opened 2019-01-07
Fix KeySCheme => KeyScheme

skinkie:


Aurige:
This is logical, however as said in my last comment, it changes a tag name (backward compatibility issue), therefore we need to inform the group about it and be prepared to have a solution (substitution group ?) in case it is requested !

PR #42, opened 2019-01-07
Moved canalBarge transport submode from air to water

seime:
Moved from AirSubmodeEnumeration to WaterSubmodeEnumeration as it is clearly not an air submode


skinkie:
Looks good.

PR #49, opened 2019-02-22
Constraint bugs

Aurige:
Addition of _**_EquipmentPlace_**_ that was missing reference in the **Place_AnyVersionedKey** key constraint


skinkie:
@Aurige could you please *not* change this like description etc. in the same pull request? Hence split it in two, send a pull request for only the description. Then I'll will merge this one after that one is merged.

Aurige:
Sorry, which description are you talking about ?

skinkie:
Line 19-55

Aurige:
Ok, that's not a comment change... just a reorganisation of the layout automatically done by XML Spy (no change in the content) .... I will unset this feature for next PR

skinkie:
In that case, that requires that you would run our xmllint script on your repo, so the organisation retains the same format. See this commit: https://github.com/NeTEx-CEN/NeTEx/commit/c07c8850e009b308f23e6a3fdecd903d97d6479b

Aurige:
just tried it, but it looks that the behaviour is not consistent from one environment to the other: I ended up with all files slightly changed .... So I just unset the option in XML Spy so this will not occure any more in future PRs

skinkie:
I'll try to resolve your request later.

PR #64, opened 2019-03-05
Comment NL48

Aurige:
Add Dynamic+Print to StopPointInJourneyPattern (same as for ServiceJourney)


skinkie:
image

Everything is merged via the work of @nick-knowles, but as you can see from the documentation there is Xxxxxxx, which should be solved.

ue71603 (approved):

skinkie (approved):

PR #110, opened 2020-08-10
Change TypeOfFuelEnumeration to match official NeTEx specification

Bertware:
This commit changes the values of TypeOfFuelEnumeration in order to match the allowed FuelType values, described in table 215 (page 442) of CEN/TS 16614-1:2014.

**Note:** I am aware of the failing build, caused by an example train with fuel type "Electricity" - I will fix the example in case we agree that Electricity isn't a valid fuel type, or add Electricity back as a fuel type in case the documentation is wrong.


skinkie:
naturalGas and electricity are not in the documentation?

Bertware:
Not sure if we are allowed to share screenshots of a single table, but no, they are not:
image

Electricity is either battery or electricContact. NaturalGas is probably TPG/liquid gas/methane? I left "other" in even though it isn't specified in the standard either, but felt it was best to keep it in.

skinkie:
I want some other people to take a look here too, because I assume the documentation is _also_ incomplete.

skinkie (approved):

ue71603 (approved):

PR #117, opened 2020-08-20
Make MainPartRef non-mandatory for JourneyPart (JourneyPartGroup)

syversenkr:
When e.g. changing Train Number due to bordercrossing there is no need to divide the vehicle, but you may still want to split the journey into parts. In this use-case there is no “Main Part” thus no point in defining or referencing it.
_Minor change, fully backwards compliant_


syversenkr:
As a sidenote:
In NeTEx part 2 chapter 7.4.11.3 Coupled Journey onwards, in the UML model the MainPartRef is documented as non-mandatory [0...1], however in Table 77 – JourneyPartReferencesGroup – Group it is inconsistently documented as mandatory 1:1 (_and previous to this PR implemented as such in the XSD_ )

ue71603 (approved):

skinkie:
@ue71603 to me this feels like a hack for a _use case_. Why would you want to split a journey when crossing a border?

xsd/netex_part_2/part2_journeyTimes/netex_coupledJourney_version.xsd
```diff
@@ -210,7 +210,7 @@ of the corresponding VEHICLE TYPE. true for forward.</xsd:documentation>
 			<xsd:documentation>Reference Elements for JOURNEY PART.</xsd:documentation>
 		</xsd:annotation>
 		<xsd:sequence>
-			<xsd:element name="MainPartRef" type="JourneyPartRefStructure">
+			<xsd:element name="MainPartRef" type="JourneyPartCoupleRefStructure" minOccurs="0">
```
skinkie, line 213:
```suggestion
			<xsd:element name="MainPartRef" type="JourneyPartRefStructure" minOccurs="0">
```

Aurige: I agree with this suggestion

Aurige: and accept the PR provided that the suggestion is validated

skinkie: Do you want me to commit it, and continue? I am actually inclined to apply it to next.

Aurige (approved):

skinkie:
Contains fixes as well... so I am mild...

ue71603:
@skinkie  "splitting" journeys at borders. Just as a comment. For trains this is often done as normally the operator changes at that time officially. Also tariff reasons. International tarin journeys are modeled as two parts with an JourneyMeeting. Often this is not done at the border, but at the first commercial stop abroad. But just a side note.

ue71603:
@skinkie  but what definitively needs to be modeled is the switch in the ITCS. This might be at a border or an arbitrary point. Not needed for PI directly, but the controlling ITCS is the source of the information.

PR #126, opened 2020-11-03
Add support for NAND, NOR, XNOR as a LogicalOperation. Also renamed B…

seime:
Also renamed `BooleanOperatorEnumeration` to `LogicalOperationEnumeration` to better reflect actual use

https://en.wikipedia.org/wiki/Logic_gate

### DIFF
--- a/xsd/netex_framework/netex_responsibility/netex_validityCondition_support.xsd
+++ b/xsd/netex_framework/netex_responsibility/netex_validityCondition_support.xsd
@@ -95,9 +95,9 @@ Rail transport, Roads and Road transport
 		</xsd:complexContent>
 	</xsd:complexType>
 	<!-- ======================================================================= -->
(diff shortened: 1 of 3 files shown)


syversenkr:
Supporting the enumeration name change, to clearer distinguish logical operation and its "gates" (even if these boolean operators) form the xsd boolean data type ("true"/"false") and unary/binary (boolean) operatos ('!', '&', "&&", '|', "||", '^' etc.)

seime:
(this change is backwards compatible, the XML structure remains the same)

Aurige:
in fact that may have some consequences on code generated from the XSD, but that's a small impact, especially since they are still very few implementation of this part of NeTEx ... so Ok for me

skinkie (approved):

ue71603 (approved):

PR #136, opened 2020-12-09
Allow referencing to Timeband, UicOperatingPeriod and ServiceCalendar

syversenkr:
… in the temporal validity parameters assignment; required to validate/consume particular Norwegian ticket types

### DIFF
--- a/xsd/netex_framework/netex_reusableComponents/netex_serviceCalendar_support.xsd
+++ b/xsd/netex_framework/netex_reusableComponents/netex_serviceCalendar_support.xsd
@@ -15,6 +15,9 @@
 				<Date>
 					<Modified>2014-03-31</Modified>
 				</Date>
+				<Date>
+					<Modified>2020-12-08</Modified>FIX - Add UicOperatingPeriodRef
+				</Date>
 				<Description>
 					<p>NeTEx - Network Exchange. This subschema defines  SERVICE CALENDAR identifier  types.</p>
 				</Description>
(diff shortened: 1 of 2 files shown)


skinkie:
@syversenkr are you going to provide a Change Request for the standard?

syversenkr:
@skinkie No problem to do that if imperative.
@Aurige But do we currently have a proper process in place? I.e. is there any follow up on the published CR documents on Basecamp?

syversenkr:
Additionally, there are a pipeline of issues with UicOperatingPeriod. Have the full details in the Entur "Known Errors" log, but as a brief summary:
* Not defined in the spec (NeTEx part 1 - 7.7.5.5 Service Calendar)
* Part 4 (profile) 7.13.9 UicOperatingPeriod "defines" it as an OperatingPeriod subtype but it is implemented in the schema as a DataManagedObject, thus the requirement for a separate reference type

Aurige:
@syversenkr yes please send it to Data4PT ... @nick-knowles is in charge of linking the requests we receive to starndard updates.
We can also open a new dedicated folder in NeTEx Basecamp projetc.... that could be usefull

syversenkr:
@Aurige 
I second your proposal to have a new dedicated folder, as well as establishing a formal process and assign Data4PT resources (if anyone outside of the group present here on GitHub, that is) to manage this workflow, to ensure CRs are followed up at least on a semi-regular basis.

Aurige:
I reorganised a bit the CR folder. Here is the place for new ones https://3.basecamp.com/3256016/buckets/2570434/vaults/3283607012
Also we may be abale to manage some in the context of the New Modes work, if relevant

syversenkr:
Solution has been discussed with @nick-knowles 
This PR is replaced by #148

PR #138, opened 2020-12-15
Allow OperatingPeriodRef to reference to UicOperatingPeriod

skinkie:
EPIP profile uses UicOperatingPeriod, sadly we cannot use the current XSD because it does not allow OperatingPeriodRef to reference UicOperatingPeriod.


Aurige (approved):

Aurige:
It is possible to use UicOperatingPeriod via ValidityConditions, but this is obviously a nicer way

skinkie:
> It is possible to use UicOperatingPeriod via ValidityConditions, but this is obviously a nicer way

If you take a look at the Luxembourg exports, they are doing it in this way.

syversenkr (commented):
Supported!
Ref. comments in #148

PR #146, opened 2021-01-29
Need to specify more than one ActivationMeans

seime:
Currently it is only possible to define a single activation mean for a usage validity period.

This PR adds support for multiple values

This PR has been discussed with @nick-knowles


skinkie:
Don't we want to group these kind of elements?

Aurige:
fine for me (@skinkie it is a list of enumertation here, so a group of things does not fit)

skinkie:
DayTypes are a list of enumerations too right?

ue71603 (approved):

skinkie (approved):

PR #168, opened 2021-04-25
Fix AccessEquipmentRef

### LINKED ISSUE #167: [FIX] AccessEquipmentRef incorrently subsitutes EquipmentRef
AccessEquipment subsitutes InstalledEquipment, allowing a StaircaseEquipment or LiftEquipment to be added to a SitePathLink. Incorrectly, an StaircaseEquipmentRef may not be used due to EquipmentRef is subsituted instead of InstalledEquipmentRef.

image

skinkie:
Fix #167

(see issue, for documentation)


Aurige (approved):

Aurige:
I approved this fix since it makes sense, but it may have some consequences on existing datasets (but quite unlikely since Equipment or not widely used yet)

syversenkr:
Hopefully I am not missing something here, but from what I can see in the XSD this bug fix will only allow for references previously _not_ part of the placeEquipments refenece substitution (InstalledEquipmentRef), i.e. the AccessEquipmentRef subtypes. The element types themselves where already correctly of InstalledEquipment substitution type.

On the other hand: In order to align with the UML/spec, it seems these references should _more appropriately_ have been of a PlaceEquipment(Ref) substitution group.

syversenkr:
equipment-highlight

On the other hand, both the InstalledEquipment type/ref and PlaceEquipment type/ref are abstract conceptual classes of super Equipment type/ref, and this has no practical implication other than adding an extra level of dummy element/reference in the XSD.

syversenkr (approved):
See previous comments, as the abstract substitution class has no technical implication on the XSD

skinkie:
@syversenkr I did not find a PlaceEquipmentStructure did you?,

syversenkr:
No, it was not yet added. A technical shortcut when implementing, I presume. 

I can add if we want to reflect this additional layer.

skinkie:
> No, it was not yet added. A technical shortcut when implementing, I presume.
> 
> I can add if we want to reflect this additional layer.

I am checking SiteEquipmentRef and SignEquipmentRef which have exactly the same problem.

<xsd:element name="SiteEquipmentRef" type="SiteEquipmentRefStructure" substitutionGroup="EquipmentRef">
<xsd:element name="SignEquipmentRef" type="EquipmentRefStructure" substitutionGroup="EquipmentRef">

So the question is, should we create the PlaceEquipment as substitution?

syversenkr:
Both results in the same XML, so no technical impact if wrapping in one or two abstract classes. More a discussion if worth the effort, i.e. should the XSD implement the spec _technically_ or as close to the model as possible.

However, the additional incorrect references with substitution group Equipment should be fixed as part of this commit.

xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_version.xsd
```diff
@@ -128,7 +128,7 @@ Rail transport, Roads and Road transport
 		</xsd:sequence>
 	</xsd:group>
 	<!-- ====ACCESS=================================================================== -->
-	<xsd:element name="AccessEquipment" type="AccessEquipment_VersionStructure" substitutionGroup="InstalledEquipment">
+	<xsd:element name="AccessEquipment" type="AccessEquipment_VersionStructure" abstract="true" substitutionGroup="InstalledEquipment">
```
skinkie, line 131: What does this 'abstract' mean?

syversenkr: ^
 |

syversenkr: Before fix:

AccessEquipment-list-element

=> Validation error "The type definition cannot be abstract for element AccessEquipment."

syversenkr:
The question being: Should we create the PlaceEquipment as substitution to align with the UML model? Or keep the simpler implementation, skipping abstract types when it does not have a practical implication?

Aurige:
> The question being: Should we create the PlaceEquipment as substitution to align with the UML model? Or keep the simpler implementation, skipping abstract types when it does not have a practical implication?

I would keep it simple

skinkie:
I think the only practical implementation this will have is if we are to be generating UML diagrams from the XSD, it must be modeled.

skinkie:
I suggest we are going to do PlantUML, hence do it via the long way.

nick-knowles:
The AccessEquipment should also be abstract (not just the AccessEquipmentRef)

ue71603:
@skinkie  This is still blocked?

skinkie:
> @skinkie This is still blocked?

Not blocked, but a suggestion what a solution would be, would be appreciated. I am happy to add abstract=true to all the blue classes.

Aurige:
yes according to the UML they look abstract

xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_support.xsd
```diff
@@ -69,7 +69,7 @@ Rail transport, Roads and Road transport
 		</xsd:annotation>
 		<xsd:restriction base="InstalledEquipmentIdType"/>
 	</xsd:simpleType>
-	<xsd:element name="AccessEquipmentRef" type="AccessEquipmentRefStructure" substitutionGroup="EquipmentRef">
```
ue71603, line 72: This one is covered by #293

xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_support.xsd
```diff
@@ -79,10 +79,10 @@ Rail transport, Roads and Road transport
 			<xsd:documentation>Type for a reference to an ACCESS EQUIPMENT.</xsd:documentation>
 		</xsd:annotation>
 		<xsd:simpleContent>
-			<xsd:restriction base="EquipmentRefStructure">
```
ue71603, line 82: This is done in #293

ue71603, line 82: This is done in #293

xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_version.xsd
```diff
@@ -128,7 +128,7 @@ Rail transport, Roads and Road transport
 		</xsd:sequence>
 	</xsd:group>
 	<!-- ====ACCESS=================================================================== -->
-	<xsd:element name="AccessEquipment" type="AccessEquipment_VersionStructure" substitutionGroup="InstalledEquipment">
```
ue71603, line 131: already done in master

ue71603, line 131: already done in master

xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_support.xsd
```diff
@@ -79,10 +79,10 @@ Rail transport, Roads and Road transport
...
+			<xsd:restriction base="InstalledEquipmentRefStructure">
 				<xsd:attribute name="ref" type="AccessEquipmentIdType" use="required">
 					<xsd:annotation>
-						<xsd:documentation>Identifier of a TICKETING EQUIPMENT.</xsd:documentation>
```
ue71603, line 85: wrong in master. This some documentation change should be done.

ue71603, line 85: wrong in master. This some documentation change should be done.

xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_support.xsd
ue71603, line 72: This is new. It currently works with EquipmentRef (as this is the more generalized class). However, to be very in line, whe should probably keep this change. How pedantic do we want to be.

ue71603, line 72: This is new. It currently works with EquipmentRef (as this is the more generalized class). However, to be very in line, whe should probably keep this change. How pedantic do we want to be.

ue71603 (changes requested):
Changing  <xsd:element name="AccessEquipmentRef" type="AccessEquipmentRefStructure" substitutionGroup="EquipmentRef"> we should consider. Correcting the documentation TICKETING to ACCESS. The rest should be thrown out.

xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_support.xsd
ue71603: https://github.com/NeTEx-CEN/NeTEx/pull/293/files contains an example. Therefore it should break master, if we have a problem there.

PR #173, opened 2021-05-05
BackgroundColour capitalization

### LINKED ISSUE #172: BackgroundColour wrongly added to XSD
image

https://github.com/NeTEx-CEN/NeTEx/blob/master/xsd/netex_framework/netex_utility/netex_utility_types.xsd#L261

skinkie:
Fix #172

### DIFF
--- a/xsd/netex_framework/netex_utility/netex_utility_types.xsd
+++ b/xsd/netex_framework/netex_utility/netex_utility_types.xsd
@@ -258,7 +258,7 @@ Rail transport, Roads and Road transport
 					- Pantone (be aware that Pantone is proprietary), etc.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
-			<xsd:element name="BackGroundColour" type="ColourValueType" minOccurs="0">
+			<xsd:element name="BackgroundColour" type="ColourValueType" minOccurs="0">
 				<xsd:annotation>
 					<xsd:documentation>Default RGB background colour value for text and graphics associated with ENTITY - e.g. of a LINE.</xsd:documentation>
 				</xsd:annotation>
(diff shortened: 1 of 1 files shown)


Aurige:
I'm afraid that this attribute is already in use in several case.... fixing this type will be a backgroung compatibility issue for several implementers (so I don't think wen can removed the mistyped element)

skinkie:
> I'm afraid that this attribute is already in use in several case.... fixing this type will be a backgroung compatibility issue for several implementers (so I don't think wen can removed the mistyped element)

Then they have to update. Since they did not implement the CEN Documentation which clearly mentions the lower-G.

syversenkr:
My assumption has been that those needing to maintain using invalid enumerations (e.g. obvious typing errors) without reporting these, rely on previous versions of the XSD anyway?

If still needing to maintain this deviation from the spec for backwards compatibility, per previous discussions, please add this as a new commit to this pull request. But from my point of view, we do satisfy backwards compatibility in this particular case through previously published XSD versions.

syversenkr (approved):
See my previous comment.

Additionally, if further commits are made to this PR (e.g. keeping also the deprecated element name with incorrect spelling), one should consider grouping the common elements of PresentationStructure and PrintPresentationStructure instead of duplicating

Aurige:
@syversenkr even if an organisation can always stay on the previous XSD, the difficulty is that NeTEx is an exchange protocol: for example in Ile de France we have 87 operators using NeTEx, if one migrates to the new XSD he will not be able to exchange with the others because his file will trigger validation issue (same thing for the ones he receives). So suche a change requires that the 87 operator migrate at once, which is pretty much impossible.
This  said, after checking again I finally  didn't find any operational use of this attribute in France (from the ones I know), so I'm going to validate the PR (I hope there are no usage elsewhere, but since nobody else raised hand about it this morning about it, it should be Ok ... fingers crossed).

Aurige (approved):

PR #177, opened 2021-05-06
PathNavigationGroup: GradientType

skinkie:
image
image


Aurige (approved):
possible backward compatibility issue, but I'm not aware of existing implementations... also I'm aware of several coming very soon

PR #179, opened 2021-06-09
Add allowed value 'proposed'

syversenkr:
NeTEx part 1 chapter _7.3.4.4.2.3 VersionStatus – Allowed Values_ documents that the VersionStatusEnumeration should include 'proposed' ("_This version is comprehensive but not yet validated._")

NB: Implemented value "other" is **not** currently documented in the spec.


skinkie:
I don't think this is what the standard means with "new", "deleted", "delta". It is a technical meaning, proposed is something different, it is _context_, not a version management option.

skinkie (approved):

Aurige (approved):

PR #183, opened 2021-07-06
New modes pooler profile revisions

nick-knowles:
Minor update for NewModes review .  Add luggage allowed to pooler profile and , revise example to show use of luggage allowance


Aurige:
I'm afraid this PR is routed to Master instead of NewModes ... so it looks like the first huge PR ! Can you reroute it to the NewMode branch @nick-knowles ?

skinkie:
You can do it yourself, just press "Edit".

Aurige (commented):
@nick-knowles the netex_trip_support.xsd  is new in NeTEx (I don't find it in previous versions) ? and it is just the "_support" file, but without the corresponding "_version" .... shouldn't we skip the trip references for now ?

PR #203, opened 2021-07-20
Fix VehicleAccessFacilityListOfEnumerations to VehicleAccessFacilityEnumeration

skinkie:


Aurige (approved):

Aurige:
Needs to be also applied to New Modes

skinkie:
@Aurige Merging should be done for that to happen automatically.

PR #220, opened 2021-09-01
NewModes - add PolicyUrl for PoolingService and SharingServiceNew modes cc add policy url

nick-knowles:
Country Comment add PolicyUrl for Vehicle PoolingService and SharingServiceNew


skinkie:
Please target the correct branch :)

PR #225, opened 2021-09-07
Lets see how a downstream change works!

skinkie:


skinkie:
@nick-knowles @Aurige you may press merge :)

Aurige:
It looks Ok :-)

Aurige:
Is it something we need to do manually for each update on Master (until New Modes is merged to master) or is there a kind of automation ?

skinkie:
I don't know yet if it can be automated, but there is something under "Actions" that might allow it.

PR #238, opened 2021-09-16
Fix typo

frodrigo:


xsd/netex_part_1/part1_ifopt/netex_ifopt_path_version.xsd
```diff
@@ -176,7 +176,7 @@ Rail transport, Roads and Road transport
 			</xsd:element>
 			<xsd:element name="RightSideBorder" type="BorderTypeEnumeration" minOccurs="0">
 				<xsd:annotation>
-					<xsd:documentation>Type of border on the right side (n the direction of the PATH LINK).+v1.1</xsd:documentation>
+					<xsd:documentation>Type of border on the right side (in the direction of the PATH LINK).+v1.1</xsd:documentation>
```
skinkie, line 179:
```suggestion
					<xsd:documentation>Type of border on the right side (in the direction of the PATH LINK).</xsd:documentation>
```

Aurige (approved):
I think we can keep the +v1.1 since that's when in was introduced, but anyway that's ok

skinkie (approved):

PR #239, opened 2021-09-17
Fix typo and doc of SharpsDisposal

frodrigo:


skinkie:
Great catch!!

frodrigo:
> I would support an XML example with this PR.

Not sure what you are asking for.

skinkie:
You are updating the XSD due to a bug. But if we want to make sure that nobody is affected by such changes in the future, we should make an example using this element that is (automatically) validated prior to merging.

Aurige (approved):
some risks with backwarrd compatibility, but obviously a bug/typo to solve

skinkie:
@Aurige the quicker they are resolved, the less impact ;)

skinkie (approved):

PR #249, opened 2021-10-09
The remaining CR from Adrian's list

ue71603:
I think we will have to discuss it.


ue71603:
I thought we discussed that I should add it there. Or should I do master?

Aurige:
To me that's for 'Next' branch and requires a doc update

ue71603:
Did it new with more things from Joost.

trurlurl:

PR #252, opened 2021-11-04
Add VehicleMode to VehicleType (and some typos)

hennoit:
According to Transmodel (paragraph 5.6.2): _"A VEHICLE TYPE must belong to one TRANSPORT MODE. For instance, the 'bus' TRANSPORT MODE will gather standard, articulated, minibus, double-deck buses."_
But there is no way to express this in a VehicleType, both in Transmodel (paragraph 5.6.10) and in NeTEx (paragraph 7.7.11.3.2 and xsd). In the Dutch Profile we need this relationship.


skinkie:
Those typo's should have been separate...

skinkie:
I'll merge #253 first to master, then get that into next. And then do this one.

hennoit:
After the merge with "new-modes" VehicleType already contains TransportMode, so there is no need (anymore) to add VehicleMode.

PR #255, opened 2021-11-04
Allow multiple gml:Polygons associated to a Zone

hennoit:
For certain types of _Zone_ (e.g. TransportAdministrativeZone) there are situations where the area represented by the Zone is not **one** surface, but consists of multiple parts. The most common case is a region consisting of several islands.
These cases cannot be described by the current implementation of Zone, because it contains only one _gml:Polygon_. The problem can be solved by allowing multiple polygons to be associated to the Zone.

### DIFF
--- a/xsd/gml/geometryAggregates-extract-v3_2_1.xsd
+++ b/xsd/gml/geometryAggregates-extract-v3_2_1.xsd
@@ -1,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<schema xmlns:xlink="http://www.w3.org/1999/xlink" xmlns:gml="http://www.opengis.net/gml/3.2" xmlns="http://www.w3.org/2001/XMLSchema" targetNamespace="http://www.opengis.net/gml/3.2" elementFormDefault="qualified" version="3.2.1.2">
+	<annotation>
+		<appinfo source="urn:x-ogc:specification:gml:schema-xsd:geometryAggregates:3.2.1">geometryAggregates.xsd</appinfo>
+		<documentation>See ISO/DIS 19136 12.3.
+Geometric aggregates (i.e. instances of a subtype of gml:AbstractGeometricAggregateType) are arbitrary aggregations of geometry elements. They are not assumed to have any additional internal structure and are used to "collect" pieces of geometry of a specified type. Application schemas may use aggregates for features that use multiple geometric objects in their representations.
+
+GML is an OGC Standard.
+Copyright (c) 2007,2010 Open Geospatial Consortium.
+To obtain additional rights of use, visit http://www.opengeospatial.org/legal/ .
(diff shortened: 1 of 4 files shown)


Aurige (approved):

skinkie:
An objection from my side. What you want is a MultiPolygon modelling it as an unbounded Polygon is a hack.

hennoit:
> An objection from my side. What you want is a MultiPolygon modelling it as an unbounded Polygon is a hack.

Sounds reasonable, but that probably means adding extra structure(s) - which I am not familiar with - for MultiPolygon to the current GML extract in NeTEx.

skinkie:
I think a choice would do here?

skinkie:
> > I think a choice would do here?
> 
> between Polygon and MultiPolygon?

Exactly. It is backwards compatible, and the right choice for your use case.

hennoit:
> > An objection from my side. What you want is a MultiPolygon modelling it as an unbounded Polygon is a hack.
> 
> Sounds reasonable, but that probably means adding extra structure(s) - which I am not familiar with - for MultiPolygon to the current GML extract in NeTEx.

Is MultiPolygon already in the ISO (or CEN) standard describing GML? So far I can only find it in the OGC documentation (but perhaps I do not have access to the right documentation).
If not, is that a problem for NeTEx or can we add MultiPolygon anyway based on the OGC information?

Aurige:
> Is MultiPolygon already in the ISO (or CEN) standard describing GML?

it looks to be availble: http://xml.fmi.fi/namespace/meteorology/conceptual-model/meteorological-objects/2009/04/28/docindex43.html

hennoit:
> > Is MultiPolygon already in the ISO (or CEN) standard describing GML?
> 
> it looks to be availble: http://xml.fmi.fi/namespace/meteorology/conceptual-model/meteorological-objects/2009/04/28/docindex43.html

Thanks. 
MultiPolygon is deprecated in GML 3.0, but still available in the XSD for backwards compatibility until GML 3.1.1. Unfortunately it seems to have been dropped out of the XSD in GML 3.2.1 - which is the version included in NeTEx.
So perhaps we should use the MultiSurface (which can contain Polygons as its surfaceMembers) instead.

skinkie:
@hennoit thanks for researching this. MultiSurface seems to be the better choice.

Aurige:
fine for me too .. will need some documentation

skinkie:
> fine for me too .. will need some documentation

I would suggest to treat the next repository as a batch of individual prototyped and tested consequential CR's. If we would ever go for a NeTEx update, then we can use each (squashed) commit as the starting point.

hennoit:
Change Request 61:
https://3.basecamp.com/3256016/buckets/2570434/uploads/4506352750

PR #261, opened 2022-01-03
telecabine -> telecabin

ue71603:
Typo changed. Is in enum


skinkie:
@ue71603 you have targeted next, while it should target master. It is a fix (though breaking).

PR #262, opened 2022-01-03
Added OrderNumber for organising Levels

ue71603:
Added an OrderNumber that allows levels to be organised (somewhat).


skinkie:
First; this is a feature thus should be targeted to next, unless it is already documented and is technically missing.
Second; Through the standard we are using the Order attribute, is there an obvious reason to make it an element here?

PR #264, opened 2022-01-03
telecabine -> telecabin

ue71603:


skinkie:
Breaking change, but since this is a typo in an enumeration, it must be fixed asap.

xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_support.xsd
```diff
@@ -414,7 +414,7 @@ Rail transport, Roads and Road transport
 			<xsd:enumeration value="positionAtBusStop"/>
 			<xsd:enumeration value="boatGangway"/>
 			<xsd:enumeration value="ferryGangway"/>
-			<xsd:enumeration value="telecabineplatform"/>
```
skinkie, line 417: Is there any reason why the casing is not the same between the two instances? If we change this, can't we directly align it?

Aurige (approved):
not great for compatibility, but OK (shouldn't we keep a deprecated buggy value ?)

xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_support.xsd
ue71603: done

skinkie (approved):

PR #276, opened 2022-01-19
Update netex_ifopt_stopPlace_support.xsd

ue71603:
Extension of QuayTypeEnumeration for the hierarchy needed in rail

### DIFF
--- a/xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_support.xsd
+++ b/xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_support.xsd
@@ -380,7 +380,7 @@ Rail transport, Roads and Road transport
 	<xsd:simpleType name="QuayTypeEnumeration">
 		<xsd:annotation>
 			<xsd:documentation>Allowed values for QUAY Component Types. Special is the structure for rail, which allows a hierarchy. 
-			railIslandPlatform is the full platform, railPlatform is the track and railPlatformSector is a sector of such a railPlatform. </xsd:documentation>
+			railIslandPlatform is the full platform positioned between two tracks, railPlatform is a side platform facing the track and railPlatformSector is a sector of such a railPlatform. A similar hierarchy is defined for a metro station layout.</xsd:documentation>
 		</xsd:annotation>
 		<xsd:restriction base="xsd:string">
 			<xsd:enumeration value="airlineGate"/>
--- a/xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_support.xsd
+++ b/xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_support.xsd
@@ -379,11 +379,14 @@ Rail transport, Roads and Road transport
 	</xsd:simpleType>
 	<xsd:simpleType name="QuayTypeEnumeration">
 		<xsd:annotation>
-			<xsd:documentation>Allowed values for QUAY Component Types.</xsd:documentation>
+			<xsd:documentation>Allowed values for QUAY Component Types. Special is the structure for rail, which allows a hierarchy. 
+			railIslandPlatform is the full platform, railPlatform is the track and railPlatformSector is a sector of such a railPlatform. </xsd:documentation>
 		</xsd:annotation>
 		<xsd:restriction base="xsd:string">
 			<xsd:enumeration value="airlineGate"/>
-			<xsd:enumeration value="railPlatform"/>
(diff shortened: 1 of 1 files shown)


Aurige (approved):

Aurige:
so do we target Next ?

skinkie:
@Joostb61 this is what we want?

ue71603:
@Aurige  I just did it to have it somewhere. I think you and @skinkie  should resolve, where the epiap branch will be and then I redo it again :-)

Joostb61:
For a crossplatform transfers it may also be relevant to distinguish MetroIslandPlatform. What do you think?

Aurige:
@ue71603 @skinkie in fact going to "next" is Ok for me... Monday I missed the very beginning of the discussion, and directly jumped in when you were discussing about an "accessibility" branch ... but as pointed by the discussion this raises the question of where to branch it ... at the end of the day, I would feel more comfortable with just "master" for bug correction and "next" for any change/improvement

skinkie:
@Aurige especially for this 'generic' improvement next would work for me.

skinkie:
> For a crossplatform transfers it may also be relevant to distinguish MetroIslandPlatform. What do you think?

I agree.

xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_support.xsd
```diff
@@ -379,11 +379,14 @@ Rail transport, Roads and Road transport
...
+			<xsd:documentation>Allowed values for QUAY Component Types. Special is the structure for rail, which allows a hierarchy. 
+			railIslandPlatform is the full platform, railPlatform is the track and railPlatformSector is a sector of such a railPlatform. </xsd:documentation>
 		</xsd:annotation>
 		<xsd:restriction base="xsd:string">
 			<xsd:enumeration value="airlineGate"/>
-			<xsd:enumeration value="railPlatform"/>
+			<xsd:enumeration value="railPlatform"/> 
+			<xsd:enumeration value="railIslandPlaform"/>  
+			<xsd:enumeration value="railPlatformSector"/>
 			<xsd:enumeration value="metroPlatform"/>
```
skinkie, line 390:
```suggestion
			<xsd:enumeration value="metroPlatform"/>
			<xsd:enumeration value="metroIslandPlatform"/>			
```

xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_support.xsd
```diff
@@ -379,12 +379,16 @@ Rail transport, Roads and Road transport
 	</xsd:simpleType>
 	<xsd:simpleType name="QuayTypeEnumeration">
 		<xsd:annotation>
-			<xsd:documentation>Allowed values for QUAY Component Types.</xsd:documentation>
+			<xsd:documentation>Allowed values for QUAY Component Types. Special is the structure for rail, which allows a hierarchy. 
+			railIslandPlatform is the full platform, railPlatform is the track and railPlatformSector is a sector of such a railPlatform. </xsd:documentation>
```
skinkie, line 383:
```suggestion
			railIslandPlatform is the full platform positioned between two tracks, railPlatform is a side platform facing the track and railPlatformSector is a sector of such a railPlatform. A similar hierarchy is defined for a metro station layout.</xsd:documentation>
```

ue71603:
Some more comments:
Wilfried thinks that we should model it a bit differently:
* Platform is the whole thing and PlatformSide is only a part
* So we would model: RailwayPlatform having two RailwayPlatformSide and for each RailwayPlatformSide 0...n RailwayPlatformSector

Alex thinks:
* Metro should be subdivided the same way. 
* A short trian may have a special sector: image
* Eventually even bus and tram need to have the subdivsion. Espacially for cases where there is a part that is higher (or lower) for level access (Alex provided an image). This would be a sector, even if there are otherwise none:
image


I suggest to:
* rename
* change the explanation
* add MetroPlatform, MetroPlatformSide and MetroPlatformSector

Should I also add
* HandicappedLevelBoardingSector

skinkie:
The point is, every PlatformSide is a Platform. But not every Platform is an IslandPlatform. I think it is too complex to introduce the Platform -> PlatformSide hierarchy, and it creates a backwards compatibility issue, which my proposal does not have.

ue71603:
Updated from the discussion with backward compatibility

klement-MENTZ (approved):

PR #278, opened 2022-01-21
Update netex_facility_support.xsd

ue71603:
TSI-PRM IoA has a "Dependent on the train" value for PRM assistance. The current set of AssistanceAvailability does not really model that. It only models availableAtCertainTimes. So I thought this would be a good update. Also Adrian and I thought that the "train" is probably more Journey than Vehicle.

### DIFF
--- a/xsd/netex_framework/netex_reusableComponents/netex_facility_support.xsd
+++ b/xsd/netex_framework/netex_reusableComponents/netex_facility_support.xsd
@@ -281,14 +281,19 @@ Rail transport, Roads and Road transport
 			</xsd:enumeration>
 			<xsd:enumeration value="availableIfBooked">
 				<xsd:annotation>
-					<xsd:documentation>Assistance is  available if booked.</xsd:documentation>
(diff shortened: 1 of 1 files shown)


Aurige:
Ok for me, but how do you then know which journeys are Ok  ? one way to do it may be through the ConditionedObjectRef in the ValidityCondition

ue71603:
From TSI PRM we don't know it. From NeTEx this will be done by Alex Klement in the Stop-Vehicle interaction. Really one needs to analyse a lot of factors. So to really have a ValidityCondition there would be a sort of shortcut. I think the trip planners should be able to process the TrainComponents etc if they want to have this detailed a view on what is possible and what not. Or do you think the operators can provide the ValidityCondition?

Aurige:
my point is that only having **availableDependentOnJourney** without any relation to Journeys sounds a bit like **sometimes**

ue71603:
If you prefer we use the availableAtCertainTimes which is exactly that value and we skip the additional value and I adapt the mapping NeTEx-TSI PRM to say that availableAtCertainTimes is the equivalent. Is ok for me....

PR #280, opened 2022-01-24
Updating additional relevant user categories

ue71603:
women and families are possibly also relevant additional ParkingUserTypes


skinkie:
Agreed. We have parking signs for them.

skinkie:
Is taxi, sharedVehicle, pickup/dropoff stuff also desirable here?

ue71603:
I think this is really the hard core parking that is meant. But perhaps @Aurige  has an opinion here.
(all from the OeBB example).

Aurige:
@skinkie we have dedicated places in new modes for  sharedVehicle, pickup/dropoff
Do we really have use cases for women and families for parking users ? some parking only for women !!!!??

skinkie:
> @skinkie we have dedicated places in new modes for sharedVehicle, pickup/dropoff Do we really have use cases for women and families for parking users ?  some parking only for women !!!!??

These are typically bigger places within a garage without any obstacles near by.   Near stores we also have them often.

Aurige:
we have VehicleMeetinPlace for that in New modes

skinkie:
It is still a different meaning. The family/woman spots are about 1.5x the width for parking.

Aurige:
you mean that larger places are said to be for woman ? isn't it a bit sexist ? why not for elders ? or beginners ? 

 _(sorry for the MeetingPlace answer... I mixed 2 conversations)_

skinkie:
Legislation in The Netherlands and Germany is there to help people, not to be 'woke'. We have coloured zebra lanes too, doesn't mean they are exclusively used for the LGBTIQ+ community...

ue71603:
It is a thing that exists: https://de.wikipedia.org/wiki/Frauenparkplatz . So that's why I added the Type

Aurige:
So, I learn something today (I just read several articles about it). If I can understand the security concern (as there are some carriage reserved for women in some countries... which by the way also tells how bad men can be...) I really can't understand why they painted them pink and made them larger than other places !! 
Anyway since it is in some laws in EU, I guess we need to add these values.

Issue #286, opened 2022-01-26
sequence, sequence


skinkie:
I just notice that OnwardTimingLink_DerivedViewStructure  has a sequence, sequence, is this a bug?

https://github.com/NeTEx-CEN/NeTEx/pull/285/files#diff-70b0173f5ac6c906d3d98d0b4981773743227e970909ebcee169eb217cafa45fR551

Aurige:
I just checked: in this specific case you can remove the nested `<sequence>` (it does not change the content and would be nicer modeling)... but you cant' to it everywhere !
This is the XML Spy view as it is now 
![image](https://user-images.githubusercontent.com/3328999/151188687-8a52fcae-eee5-4926-9fc3-33db9a584b7b.png)

And the one where one level of nested `<sequence>` is removed
![image](https://user-images.githubusercontent.com/3328999/151188806-952cfc8d-24d2-4360-ba08-dd977f6ed4cc.png)

I also checked that there was not other XSD Restriction with _base="OnwardTimingLinkView"_  elsewhere

So no pb here... but this depends on what structure the XSD Restriction is based on: if the original version have nested   `<sequence>` (with elements in between, for example due to the use of some Groups, etc.) and you want to keep some elements of the lower level of nesting but not the others, then you will end up with such nested <sequence> without any element in-between

skinkie:
@Aurige can you functionally define where the duplicate should stay? I already defined a nice pcregrep query to find all the variants. If we can make the schema more simple in this way, I would love to take that opportunity.

Aurige:
If you extract all the empty nesting, then we can check them... but I don't see how to avoid a manual check (one by one) ... but if this is not hundreds of them, it is doable

skinkie:
> If you extract all the empty nesting, then we can check them... but I don't see how to avoid a manual check (one by one) ... but if this is not hundreds of them, it is doable

There are recurring patterns. So if you state this pattern cannot be simplified, then we don't even try on that pattern.

The following pattern EntityInVersionGroup, DataManagedObjectGroup, ... is quite common. If you have a hint why this one can or can't be replaced, please let me know.

```xml
<xsd:element name="RailwayJunction" abstract="false" substitutionGroup="InfrastructurePoint">
        <xsd:annotation>
                <xsd:documentation>A type of INFRASTRUCTURE POINT used to describe a RAILWAY network.</xsd:documentation>
        </xsd:annotation>
        <xsd:complexType>
[...]

PR #287, opened 2022-01-26
Remove redundant sequence

skinkie:
For discussion: #286


ue71603:
@skinkie  How do we progress?

skinkie:
@ue71603 @Aurige mentioned this needs to be checked for every instance where this happens in EA. I have obviously sent him every instance...

ue71603 (approved):
There was only one to be found? Then ok

skinkie:
> There was only one to be found? Then ok

No, no, no... hunderds... I'll mail ;-)

ue71603:
@skinkie  @Aurige 
If this should be done:
(1) Say if we do it?
(2) List to be provided here
(3) @trurlurl may start working on it.

Aurige:
It was discussed in one of the meetings, and the decision was to correct this kind of unnecessary sequence when we work on something related (by opportunity): it is too much work (and time) to check all of them one by one, and it does not harm.

trurlurl (approved):

Aurige (approved):

PR #291, opened 2022-02-01
Proposal for URI with an object reference.

skinkie:
As discussed: https://3.basecamp.com/3256016/buckets/2570434/messages/4588900379

We currently don't specify what a reader should have available first to process a file, which would require to process all files first. This is a proposal and example how this could work.

### DIFF
--- a/examples/functions/relations/test-01-main.xml
+++ b/examples/functions/relations/test-01-main.xml
@@ -1,0 +1,127 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no" ?>
+<PublicationDelivery xmlns="http://www.netex.org.uk/netex" version="ntx:1.1">
+
+  <PublicationTimestamp>2022-01-11T16:04:26</PublicationTimestamp>
+
+  <ParticipantRef>DK</ParticipantRef>
+
+  <PublicationRequest>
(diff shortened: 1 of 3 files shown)


Aurige:
Thinking a little more about it: shouldn't we reuse (or specialise) the explicit codespace mechanism we already have ? This will avoid repeating the same URI over and over. The specialised codespace could also contain a reference to the containting frame (as you suggested at some point) and may be to its version.

skinkie:
> Thinking a little more about it: shouldn't we reuse (or specialise) the explicit codespace mechanism we already have ? This will avoid repeating the same URI over and over. The specialised codespace could also contain a reference to the containting frame (as you suggested at some point) and may be to its version.

You would want to propose a code space per individual dataset?

Aurige:
the idea would be to complement it if necessary and of course have a direct reference to the file to avoid any unstable id of course. in addition to the file itself, this would allow to have Frame/Version/Description information and all with one single compact code (instead of repeating the full file URI over and over).

ue71603 (approved):

ue71603:
@trurlurl  @Aurige  If you can approve, then we can merge this one.

Aurige (approved):

PR #295, opened 2022-02-16
EPIAP: PlatformHeight introduced on StopPlaceSpace

skinkie:


Aurige:
Shouldn't it be on Quays ? StopPlaceSpace can be waiting rooms, corridors, etc. I since it is an enhancement it should go to "next"

skinkie:
@Aurige we already had this discussion. We want to have in Quays, BoardingPositions, etc. If there should be introduced a different substitiongroup to facilitate enhancements that only relate to things along side of the track, that should be modelled first upstream.

Aurige:
Yes I understood that this information needs to be shared between Quays and BoardingPositions, but that's not a reason to spread it to other unrelated object. I would recommend doing an XSD Group , as it is most usually done in NeTEx, and use that group in both objects. Another to go could be to detail the restriction on AccessSpace and VehicleStoppingPlace, but that looks a bit unnatural too me.

ue71603:
May I ask you @skinkie  and @Aurige to resolve this? If modeling in a group occurs, who does it?

ue71603:
Is replaced by: https://github.com/NeTEx-CEN/NeTEx/pull/298

PR #296, opened 2022-02-16
EPIAP: edgetotrackcenterdistance

skinkie:
Distance from centerline of the track towards the platform edge. In combination with the VehicleType half width, the gap between the platform and the vehicle floor can be calculated for vehicles without a slidingStep.


Aurige:
Same comment as for [#295](https://github.com/NeTEx-CEN/NeTEx/pull/295)

ue71603:
If problematic, we can have the discussion on Monday. E.g. at the end as modeling does not interest all present.

ue71603:
Is replaced by: https://github.com/NeTEx-CEN/NeTEx/pull/298

PR #298, opened 2022-02-21
EPIAP quay boarding positions enhancement

Aurige:
Replaces #296 / #295
Adds **EdgeToTrackCenterDistance**
Distance from centerline of the track towards the platform edge. In combination with the VehicleType half width, the gap between the platform and the vehicle floor can be calculated for vehicles without a slidingStep.

Adds **PlatformHeight**
Height of the platform relative to the ground (bus) or the rail track.

This is a way to do it using a group used by quays and  boarding position (only)


Aurige:
Note : we need to wait for Stefan Jugelt's input to finalise it.

ue71603 (approved):

skinkie (approved):

PR #301, opened 2022-03-15
EPIAP: Add raised curb mobility facility enumeration

skinkie:
This has impact on other standards such as SIRI, and it is unlikely this is in Transmodel.


skinkie:
@nick-knowles should it be KERB or CURB?

ue71603 (approved):
Ok, when Nick says, it is the correct term.

Aurige (approved):
I approve it because I didn't find a better place for it, but a raisedCurb is not really a facility, as opposed to other values in the enum

PR #313, opened 2022-04-08
The connection exists in the documentation, but was not done in the xsd

ue71603:
NeTEx part 1, figure 628


klement-MENTZ (approved):

xsd/netex_part_1/part1_tacticalPlanning/netex_passengerInformationEquipment_version.xsd
```diff
@@ -243,10 +242,15 @@ LOGICAL DISPLAY corresponds to a SIRI STOP MONITORING point.</xsd:documentation>
 			<xsd:element ref="LogicalDisplayRef" minOccurs="0"/>
 			<xsd:element ref="StopPlaceRef" minOccurs="0"/>
 			<xsd:element ref="SiteComponentRef" minOccurs="0"/>
+			<xsd:element name="PassengerInformationEquipmentList" type="PassengerInformationEquipmentEnumeration" minOccurs="0">
```
duexw, line 245: 1) Do you really want to allow several values? If yes, then it would be better to have a ListOfPassengerInformationEquipment like at other places where we have several values.
Eg. instead of 
<PassengerInformationEquipmentList>DepartureMonitor</PassengerInformationEquipmentList>
<PassengerInformationEquipmentList>AudioAnnouncement</PassengerInformationEquipmentList>
you would have
<PassengerInformationEquipmentList>DepartureMonitor AudioAnnouncement</PassengerInformationEquipmentList>

duexw: Oh I see the comments box removes the xml tags

ue71603: I am open to both. @Aurige should I add only one Enum?

Aurige:
This is also different from what is currently in the documentation.... so it should be an enhancement targeting "next" branch

ue71603:
Because of the list? Otherwise I think it is not different.

Aurige:
@ue71603 yes this list is not in the current Part 1 doc, so we will need to update the document accordingly (attribute table and XSD figure, as a minimum)

ue71603:
@skinkie @Aurige @duexw  : do it without a list in master or with list in next? When decided, it will update it.

Aurige:
Looking deeper at it, I guess that the figure you are referring to in Part 1 is 773 in last release from CEN (not 628 which  is _MoneyService – XSD_ !!)
Also that's true that there is is mismatch in between the XSD and Physical UML, so at that point my suggestion would be:
- qualify it as a bug (already done)
- integrate the **List version** in XSD (can be master)
- but also note that a document update is needed (so keep the label) since the attribute table and XSD figure in the document doesn't contain this list. This document update will be done during the incoming revision.

xsd/netex_part_1/part1_tacticalPlanning/netex_passengerInformationEquipment_version.xsd
```diff
@@ -257,6 +257,8 @@ LOGICAL DISPLAY corresponds to a SIRI STOP MONITORING point.</xsd:documentation>
 		</xsd:sequence>
 	</xsd:group>
 	<!-- ======================================================================= -->
+
```
skinkie, line 260:
```suggestion
```

xsd/netex_part_1/part1_tacticalPlanning/netex_passengerInformationEquipment_version.xsd
```diff
@@ -257,6 +257,8 @@ LOGICAL DISPLAY corresponds to a SIRI STOP MONITORING point.</xsd:documentation>
...
+
```
skinkie, line 261:
```suggestion
```

skinkie:
Can't edit it, so you have to resolve it yourself ;-)

skinkie (approved):
Nice.

trurlurl:

PR #145, opened 2021-01-29
Added missing TransportMode field to accompany existing TransportSubmode

seime:
TransportMode and TransportSubmode usually come in pairs several places in NeTEx. 
In parameter assignments/validityParameters it is missing, instead there is a VehicleMode.

This PR adds TransportMode to accompany TransportSubmode

The PR has been discussed with @nick-knowles that will verify that it aligns well with other parts of Netex.

PR #330, opened 2022-04-19
Merge validityparameters

skinkie:
Fix #145


skinkie:
I guess I need to fight harder.

Aurige:
Modes of transport and vehicle modes are very close ... that looks weird to have both ... furthermore one is a list the other is a single value
I don't understand the need for this addition: **OrganisationValidityParametersGroup** is in  **ScopingValidityParametersGroup** which already has a **ModeValidityParametersGroup** for this purpose

ue71603 (commented):
image

VehicleModes exists now in ModeValiditdyParametersGroup and OrganisationValidityParametersGroup that results in an ambiguous model ==> remove  VehicleModes from OrganiationValidityParametersGroup.
image
image


I think we should drop the whole PR.

PR #331, opened 2022-04-19
Travis to include newMOdes

ue71603:


ue71603:
..and it fails...

examples/functions/newModes/IXSI Service 1 - Static Data.xml
```diff
@@ -113,7 +113,7 @@
 							<proposedByServices>
 								<OnlineServiceRef version="any" ref="GE:OnlineServiceRef:12"/>
 							</proposedByServices>
-							<VehicleSharingOperationRef version="any" ref="GE:VehicleSharingOperation:12"/> <!--Mandatory-->
+							<VehicleSharingRef version="any" ref="GE:VehicleSharingOperation:12"/> <!--Mandatory-->
```
skinkie, line 116: I don't think it should refer to an VehicleSharingOperation if it is not an VehicleSharingOperation itself, and it it would be than it would require something like nameOfRefClass.

examples/functions/newModes/IXSI Service 1 - Static Data.xml
```diff
@@ -129,10 +129,10 @@
 							<LogInRequired>true</LogInRequired>
 						</OnlineService>
 						
-						<VehicleSharingOperation version="any" id="GE:VehicleSharingOperation:12">
+						<VehicleSharing version="any" id="GE:VehicleSharingOperation:12">
```
skinkie, line 132: Change the id.

examples/functions/newModes/IXSI Service 1 - Static Data.xml
ue71603: changed

skinkie (approved):

Aurige (approved):

Issue #32, opened 2018-11-06
TimingPoint missing PrivateCode


skinkie:
Could someone validate that TimingPoint at this moment is missing PrivateCode?

ue71603:
I can't see a PrivateCode. There is PointNumber, however. Which might be used for this. But in this file also TiminLink
Strangly LinkSequence then has a PrivateCode. Also ServiePattern. To me it also seems a bit arbitrary, but I didn't miss it up until now.
@Aurige  @duexw : Do you think we must work on PrivateCode in more detail or not?

skinkie:
I think the private code should be moved up. This required a breaking schema change.

Aurige:
I don't miss it on my side (and never got request for it). 
If moving up PrivateCode may look nice, it's big schema change and will break compatibility for all users of already existing PrivateCode

skinkie:
> I don't miss it on my side (and never got request for it). If moving up PrivateCode may look nice, it's big schema change and will break compatibility for all users of already existing PrivateCode

@Hennoit requested it. TimingPoints are in the Dutch used in the AVL system.

PR #334, opened 2022-04-19
PrivateCode für TimingPoint

ue71603:
Fixing https://github.com/NeTEx-CEN/NeTEx/issues/32


skinkie:
My guess is that this fails because FarePoint is depending on TimingPoint and not the subsitution variant... but now you probably see why it would be benefitial to add it higher in the chain, but may break stuff. I would be interested how many of the examples would actually break.

ue71603:
grmbl. worked fine for the tests I did with XMLSpy... back to the drawing board

ue71603:
and jupp now the examples are breaking. @Aurige  Do you think that move could be done fpr next?

JohanEntur:
ℹ️ https://github.com/NeTEx-CEN/NeTEx/pull/369

trurlurl:

PR #349, opened 2022-04-26
Changed the ValidityConditionRef

ue71603:
This was a duplicate as each DataManagemedObject has one too at the beginning and it seems that almost everything is optional. Worked for me, but perhaps you can check, if it works here. Then we can fix it in mster.


skinkie:
@Aurige your call...

PR #352, opened 2022-04-29
Changed charset encoding from latin1, latin2 to UTF-8

seime:
The files contained a mix of iso-8859-1, iso-8859-2 and lowercase utf-8. 

Changed to UTF-8


seime:
Started out with that, but appears that Eclipse XML editor actually changes charset encoding when modifying the header

Aurige (approved):

PR #354, opened 2022-04-29
Formatting tools setup

seime:
Adds a command line tools to both check and verify that xsd and xml formatting is according to standard.

Standard defined in eclipsecodestyle/xml.prefs (Used in Eclipse WTW project)

Also hooked into Travis, but build will fail until all 6-700 files have been reformatted.


skinkie:
A lot of love for you.

README.md
```diff
@@ -37,6 +37,7 @@ The schema is broken down systematically into small modular files; generally for
 ### Codestyle
 Any changes to the content must be formatted according to a set of rules. The formatting rules are described using [Eclipse WTP configuration files](eclipsecodestyle/xml.prefs). 
 These may be imported into an Eclipse based editor or configured manually in your preferred editor. The configuration file is *somewhat* self-explanatory, but the important rules are:
+* Character encoding shall be `UTF-8`
```
skinkie, line 40:
```suggestion
* Character encoding must be `UTF-8`
```

seime:
Note that the reformatting of example xml files takes quite some time, but by only reformatting what has been changed since a git tag/branch makes it work rather quick.

Expecting Travis to work for 2 hours then fail until #355 has been merged

seime:
Travis bailed after 10 minutes since it is no output. Which is correct. Could extend to 120 minutes, but has to be removed after the first build as consecutive checks on only changed files will be much faster - and the default 10 minutes timeout should in fact kill the build.

ue71603 (approved):

PR #359, opened 2022-05-04
Get codestyle check in travis

skinkie:
@seime we are here now.


ue71603 (changes requested):
The changes broke all pending PR. Never do something like that again or you will fix all PR. Sorry, was not about you, but about  @seime

PR #361, opened 2022-05-13
Create epiap_test_Crossing_Equipment.xml

ue71603:


skinkie:
Can we shrink the example a bit?

ue71603:
I think I will shrink some examples. So we leave it for a week or two, then I will update two example files in one.

ue71603:
compressed.

skinkie:
Replaced by #548

PR #362, opened 2022-05-13
Stairfree Access as Limitation to be more compatible with OJP

ue71603:
...and for EP

### DIFF
--- a/xsd/netex_framework/netex_genericFramework/netex_accessibility/netex_acsb_limitations.xsd
+++ b/xsd/netex_framework/netex_genericFramework/netex_accessibility/netex_acsb_limitations.xsd
@@ -139,7 +139,7 @@ Rail transport, Roads and Road transport
 	</xsd:element>
 	<xsd:element name="StairFreeAccess" type="LimitationStatusEnumeration" default="unknown">
 		<xsd:annotation>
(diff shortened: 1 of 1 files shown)


xsd/netex_framework/netex_genericFramework/netex_accessibility/netex_acsb_limitations.xsd
```diff
@@ -136,7 +137,11 @@ Rail transport, Roads and Road transport
 			<xsd:documentation>Whether a PLACE has step free access.</xsd:documentation>
 		</xsd:annotation>
 	</xsd:element>
-	<xsd:element name="EscalatorFreeAccess" type="LimitationStatusEnumeration" default="unknown">
+	<xsd:element name="StairFreeAccess" type="LimitationStatusEnumeration" default="unknown">
+		<xsd:annotation>
+			<xsd:documentation>Whether a PLACE has stair free access.</xsd:documentation>
```
Joostb61, line 142: I suggest to add the difference with StepFree in het annotation: "Whether a PLACE has stair free access, in comparison with step free access one single step in the route is allowed".
With this addition the proposal is OK for me.

Aurige (approved):

PR #375, opened 2022-06-30
remove the duplicate ContactTypeEnumeration

ue71603:
by David


davidr1234:
This is a duplicate, because the enum exists in:
- netex_contact_support 
- netex_organisation_support
NOTE: the only difference is that the enum in netex_organisation_support **additionally** has the enum value "publicRelations"

davidr1234 (approved):
This seems OK to me, assuming the enumeration was intended for netex_contact_support and not  netex_organisation_support

skinkie:
Issue has resolved itself.

PR #388, opened 2022-08-22
Delfiplus extensions

ue71603:
* changes to ramp/hoist
* more information in stationMap
removal of an element that we could not find out, what it does (replaced with a length and width)

### DIFF
--- a/xsd/netex_framework/netex_reusableComponents/netex_equipmentVehiclePassenger_version.xsd
+++ b/xsd/netex_framework/netex_reusableComponents/netex_equipmentVehiclePassenger_version.xsd
@@ -219,13 +219,13 @@ Rail transport, Roads and Road transport
 			<xsd:element name="EquipmentLength" type="LengthType" minOccurs="0">
 				<xsd:annotation>
 					<xsd:documentation>Length of the EQUIPMENT be it hoist or ramp. When fully extended and only the part outside the VEHICLE.</xsd:documentation>
-				</xsd:annotation>			
+				</xsd:annotation>
 			</xsd:element>
 			<xsd:element name="EquipmentWidth" type="LengthType" minOccurs="0">
 				<xsd:annotation>
 					<xsd:documentation>Width of the EQUIPMENT be it hoist or ramp. When fully extended and only the part outside the VEHICLE.</xsd:documentation>
-				</xsd:annotation>			
-			</xsd:element>			
+				</xsd:annotation>
+			</xsd:element>
 			<xsd:element name="GapToPlatform" type="LengthType" minOccurs="0">
 				<xsd:annotation>
 					<xsd:documentation>Normal gap between VEHICLE and platform.</xsd:documentation>
--- a/xsd/netex_framework/netex_reusableComponents/netex_equipmentVehiclePassenger_version.xsd
+++ b/xsd/netex_framework/netex_reusableComponents/netex_equipmentVehiclePassenger_version.xsd
@@ -196,19 +196,14 @@ Rail transport, Roads and Road transport
 					<xsd:documentation>VEHICLE Hoist can be used at VEHICLE has a hoist or lift  for wheelchairs.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
-			<xsd:element name="HoistOperatingRadius" type="LengthType" minOccurs="0">
-				<xsd:annotation>
-					<xsd:documentation>Distance from VEHICLE needed to operate hoist.</xsd:documentation>
-				</xsd:annotation>
-			</xsd:element>
 			<xsd:element name="Ramp" type="xsd:boolean" minOccurs="0">
 				<xsd:annotation>
 					<xsd:documentation>Whether  a ramp may be used to access VEHICLE.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
-			<xsd:element name="RampBearingCapacity" type="WeightType" minOccurs="0">
+			<xsd:element name="BearingCapacity" type="WeightType" minOccurs="0">
 				<xsd:annotation>
-					<xsd:documentation>Maximum weight that ramp can bear.</xsd:documentation>
+					<xsd:documentation>Maximum weight that the EQUIPMENT can bear.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
 			<xsd:element name="NumberOfSteps" type="xsd:nonNegativeInteger" minOccurs="0">
@@ -221,6 +216,16 @@ Rail transport, Roads and Road transport
 					<xsd:documentation>Maximum step height to board.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
+			<xsd:element name="EquipmentLength" type="LengthType" minOccurs="0">
+				<xsd:annotation>
+					<xsd:documentation>Length of the EQUIPMENT be it hoist or ramp. When fully extended and only the part outside the VEHICLE.</xsd:documentation>
+				</xsd:annotation>			
+			</xsd:element>
+			<xsd:element name="EquipmentWidth" type="LengthType" minOccurs="0">
+				<xsd:annotation>
+					<xsd:documentation>Width of the EQUIPMENT be it hoist or ramp. When fully extended and only the part outside the VEHICLE.</xsd:documentation>
+				</xsd:annotation>			
+			</xsd:element>			
 			<xsd:element name="GapToPlatform" type="LengthType" minOccurs="0">
 				<xsd:annotation>
 					<xsd:documentation>Normal gap between VEHICLE and platform.</xsd:documentation>
--- a/xsd/netex_framework/netex_reusableComponents/netex_equipment_support.xsd
+++ b/xsd/netex_framework/netex_reusableComponents/netex_equipment_support.xsd
@@ -115,7 +115,7 @@ Rail transport, Roads and Road transport
 	</xsd:simpleType>
 	<xsd:element name="TypeOfEquipmentRef" type="TypeOfEquipmentRefStructure" substitutionGroup="TypeOfEntityRef">
 		<xsd:annotation>
-			<xsd:documentation>Reference to a TYPE OF EQUIPMENT.</xsd:documentation>
+			<xsd:documentation>Reference to a TYPE OF EQUIPMENT. Some values are defined for EPIAP "mobileRamp" for mobile Ramps, "permanentRamp" for permanent installation and "wheelchairLift" for a hoist in LiftEquipment.</xsd:documentation>
 		</xsd:annotation>
 	</xsd:element>
 	<xsd:complexType name="TypeOfEquipmentRefStructure">
--- a/xsd/netex_framework/netex_reusableComponents/netex_equipment_support.xsd
+++ b/xsd/netex_framework/netex_reusableComponents/netex_equipment_support.xsd
@@ -115,7 +115,7 @@ Rail transport, Roads and Road transport
 	</xsd:simpleType>
 	<xsd:element name="TypeOfEquipmentRef" type="TypeOfEquipmentRefStructure" substitutionGroup="TypeOfEntityRef">
 		<xsd:annotation>
-			<xsd:documentation>Reference to a TYPE OF EQUIPMENT. Some values are defined for EPIAP "mobileRamp" for mobile Ramps, "permanentRamp" for permanent installation and "wheelchairLift" for a hoist in LiftEquipment.</xsd:documentation>
+			<xsd:documentation>Reference to a TYPE OF EQUIPMENT. Some values are defined for EPIAP: "mobileRamp" for mobile Ramps, "permanentRamp" for permanent installation and "wheelchairLift" for a hoist.</xsd:documentation>
 		</xsd:annotation>
 	</xsd:element>
 	<xsd:complexType name="TypeOfEquipmentRefStructure">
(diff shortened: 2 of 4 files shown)


xsd/netex_framework/netex_reusableComponents/netex_equipmentVehiclePassenger_version.xsd
```diff
@@ -198,19 +198,14 @@ Rail transport, Roads and Road transport
 					<xsd:documentation>VEHICLE Hoist can be used at VEHICLE has a hoist or lift  for wheelchairs.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
-			<xsd:element name="HoistOperatingRadius" type="LengthType" minOccurs="0">
```
Aurige, line 201: is this replaced by something else ?

xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_support.xsd
```diff
@@ -407,6 +407,16 @@ Rail transport, Roads and Road transport
 			<xsd:enumeration value="vehicleLoadingPlace"/>
 			<xsd:enumeration value="handicappedLevelBoardingSector"/>
 			<xsd:enumeration value="multimodal"/>
+			<xsd:enumeration value="busOnStreet">
```
Aurige, line 410: That's what I was expecting from BusStop ... needs clarification

xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_support.xsd
```diff
@@ -407,6 +407,16 @@ Rail transport, Roads and Road transport
...
+				<xsd:annotation>
+					<xsd:documentation>bus stop directly on street.</xsd:documentation>
+				</xsd:annotation>
+			</xsd:enumeration>
+			<xsd:enumeration value="tramOnStreet">
```
Aurige, line 415: Same as for bus.... we have already tramStop and TramPlatform .... I was expecting tramStop to be on street.

xsd/netex_framework/netex_reusableComponents/netex_equipmentVehiclePassenger_version.xsd
```diff
@@ -198,19 +198,14 @@ Rail transport, Roads and Road transport
...
-				<xsd:annotation>
-					<xsd:documentation>Distance from VEHICLE needed to operate hoist.</xsd:documentation>
-				</xsd:annotation>
-			</xsd:element>
 			<xsd:element name="Ramp" type="xsd:boolean" minOccurs="0">
 				<xsd:annotation>
 					<xsd:documentation>Whether  a ramp may be used to access VEHICLE.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
-			<xsd:element name="RampBearingCapacity" type="WeightType" minOccurs="0">
```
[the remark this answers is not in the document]
duexw: AccessVehicleEquipment can be either a Hoist or a Ramp, but some properties are only available for either one of them. Both change proposals have the same objective: if we replace RampBearingCapacity with just BearingCapacity, then we can use it for Hoist as well. If we replace HoistOperatingRadius by an Equipment width and length property, then we can use it for both ramp and hoist. We think the risk of change is low - in Germany no one has implemented AccessEquipments yet as well.

xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_support.xsd
duexw: We refer to a situation where a bus or tram stops in the middle of the street, but there is no separate quay and/or crossing equipment. This means that passengers have to cross a lane in traffic if they want to board the vehicle. Maybe an enum name like "busOnStreetCentre" would express it more clearly.

Aurige: Ok, it was not obvious at first: it needs to be clarified so people won't use one instead of the other (and renamed_centerStreed_ instead of _onStreet_, or the _OnStreetCentre_ you proposed )

Joostb61: >   Is the essence here that boarding and alighting the vehicle take place directly on the street? If yes, then I suggest to use "busOnStreetBoarding". "tramOnStreet" is clear. It's hard for me to imagine what this would look like for a bus. For tram there is a priority rule: when a tram stops at a "tramOnStreet", the tram may not be overtaken on the right so that passengers can get on and off safely. For busses we don't have onstreetboarding in the Netherlands.

trurlurl (commented):
See inline comments

xsd/netex_framework/netex_reusableComponents/netex_equipment_support.xsd
```diff
@@ -116,7 +115,7 @@ Rail transport, Roads and Road transport
 	</xsd:simpleType>
 	<xsd:element name="TypeOfEquipmentRef" type="TypeOfEquipmentRefStructure" substitutionGroup="TypeOfEntityRef">
 		<xsd:annotation>
-			<xsd:documentation>Reference to a TYPE OF EQUIPMENT.</xsd:documentation>
+			<xsd:documentation>Reference to a TYPE OF EQUIPMENT. Some values are defined for EPIAP "mobileRamp" for mobile Ramps, "permanentRamp" for permanent installation and "wheelchairLift" for a hoist in LiftEquipment.</xsd:documentation>
```
trurlurl, line 118: wheelchairLift is defined for AccessFacilitySet, not for LiftEquipment
The mentioned tokens do not refer to real EQUIPMENTS like LiftEquipment, RampEquipment - this confuses me because they should be References to a TYPE OF EQUIPMENT, though perhaps that's perfectly correct...

xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_support.xsd
trurlurl: I have seen the following situations:
Bus stops at a place where there is no platform or pedestrian lane on the right side - passengers have to cross the road (either before or after a,rrival of the bus) for boarding. (I think this is illegal in the long run, but may exist for several years.) Almost like Joost's proposal, I would call it busBoardingOnStreet

tramBoardingOnStreet - a tram stop (which may even have a narrow platform), but on the right of the stop there is a bus lane where buses and taxis (and probably bicycles) may pass. This riskier situation might be labellled tramBoardingOnStreet (unless, perhaps, if there is a road crossing nearby - then it might be considered a regular stop).

xsd/netex_framework/netex_reusableComponents/netex_equipmentVehiclePassenger_version.xsd
ue71603: I think this was answered by @duexw

xsd/netex_framework/netex_reusableComponents/netex_equipmentVehiclePassenger_version.xsd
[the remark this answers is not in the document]
ue71603: I think this was answered by @duexw

xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_support.xsd
ue71603: added clarification.

xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_support.xsd
ue71603: changed names

xsd/netex_framework/netex_reusableComponents/netex_equipment_support.xsd
ue71603: I think this should be fine. This is not a direct reference to the type, but an enum in association.  Or do the others think it should be more aligned? @duexw @Aurige @skinkie 
We currently don't have an example. Do we need one?

xsd/netex_framework/netex_reusableComponents/netex_equipment_support.xsd
```diff
@@ -115,7 +115,7 @@ Rail transport, Roads and Road transport
...
```
trurlurl, line 118:
```suggestion
			<xsd:documentation>Reference to a TYPE OF EQUIPMENT. Some values are defined for EPIAP: "mobileRamp" for mobile Ramps, "permanentRamp" for permanent installation and "wheelchairLift" for a hoist.</xsd:documentation>
```

trurlurl: I think LiftEquipment was incorrect.

xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_support.xsd
```diff
@@ -403,6 +403,16 @@ Rail transport, Roads and Road transport
 			<xsd:enumeration value="vehicleLoadingPlace"/>
 			<xsd:enumeration value="handicappedLevelBoardingSector"/>
 			<xsd:enumeration value="multimodal"/>
+			<xsd:enumeration value="busOnStreetBoarding">
+				<xsd:annotation>
+					<xsd:documentation>bus stop directly on street. This means people waiting to board will literally standing on the street.</xsd:documentation>
```
trurlurl, line 408:
```suggestion
					<xsd:documentation>Bus stop directly on street. This means people waiting to board will literally be standing on the street.</xsd:documentation>
```

xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_support.xsd
```diff
@@ -403,6 +403,16 @@ Rail transport, Roads and Road transport
...
+				</xsd:annotation>
+			</xsd:enumeration>
+			<xsd:enumeration value="tramOnStreetBoarding">
+				<xsd:annotation>
+					<xsd:documentation>tram stop directly on street. This means people waiting to board will literally standing on the street.</xsd:documentation>
```
trurlurl, line 413:
```suggestion
					<xsd:documentation>Tram stop directly on street. This means people waiting to board will literally be standing on the street.</xsd:documentation>
```

trurlurl (approved):

nick-knowles (commented):
Why do we need BusOnStreet Boarding  and  tramOnstreet Boarding?  Is that not exactly what BusStop amd tramStop mean ?
For bus stops that are not on the street we have busBay (eg in a  bus station) and busPlatform (eg a raised platform for bus, as might be shared with tram) 

Maybe we should add more elaborate descriptions to make this clear

Or was something else meant by BusOnStreetBoarding  e.g. no pavement or an unmarked stop .  But adding separate "Stop types" for different types of  on street bus stop " pollutes the categorisation.  Would be better to have some further property that characterises an onstreet further  bus stop.    Either way it  needs to be revised as at teh moment it is either  misleading / redundant  or both.

ue71603:
@nick-knowles  This was discussed and already answered by Wilfried. We have bus stops that are in the "middle" of the street and have no direct "border". E.g. a bicycle lane or even a street lane is between the "place" where the bus stops. Passengers will cross the lane (usually a traffic light or some special markings block the traffic during boarding/alighting. That is what is meant by this. 
So it would look like this: 
image

and this is a special case. Especially PRM wise.

ue71603:
@nick-knowles also: This was voted on by the EPIAP group in several meetings. Therefore I am quite unwilling to do anything to change it. IN EPIAP there even is an image I think. If you have concrete improvements for the description, there I am open.

nick-knowles:
Thanks for the image

Apologies but I am not in  the EPIAP group so I have missed prior discussion - but my naive interpretation from the name having not seen the the discussion  (which would be typical for others and new users)  I think shows  there is a problem. 

So it is not "busOnStreetBoarding" boarding - as all busStops are "On street"  by definition - but     "busStopWithInRoadwayBoarding" ...   so at the very least the name is  highly  misleading in that it fails to make the pertinent distinction from a normal bus stop  which is also on a street - and so  and will probably be wrongly used  by naive users  for   their normal  regular "pavement boarding "  stops.

Are such  bus stops also marked by a pole or other  indicator at the side?  How do passengers know the bus or tram will stop  in the vicinity?   If that is the case then the in road boarding aspect  sounds much more like the   Transmodel / NeTex model     concept of a  QUAY +   BOARDING POSITION, attached to a STOP PLACE (rather than being  a property of theSTOP PLACE itself) ,  and what you are describing is an additional classification of  either the QUAY  type (**_QuayType_** = _BusStopInRoad_ ?)  or BOARDING POSITION (  **_BoardingPositionType_** = _positionInRoadwayAtBusStop_ ?_ ) .   .   Remember, for buses one typically has a single STOP PLACE  representing both directions e.g. "Town Hall", with two separate QUAYs  e.g. "Eastbound Town Hall" and "Westbound TownHall", representing the  QUAYs ( ie poles / bus shelters)  either side of the street ,  and if needed, one or more BOARDING POSITIONs indicating the locations for the passenger  to stand to board the bus.  

See the existing classifications
image

By  introducing  a separate stop type for certain specific combinations of QUAY and BOARDING POSITION type you are arguably   hacking the system   - and if one starts mixing up what are meant to be orthogonal concepts you soon have a combinatorial explosion and spaghetti.

Anyway, to  resolve this  at the very least there need  to be better comments in the XML for the enums  to explain the difference  between on street and on street boarding. Personally I would prefer one either renamed the stop concept  to something like inRoadway boarding, or  better still made it a classification of either the QUAY  the BOARDING POSITION  rather than the STOP PLACE, This  would allow you to specify precise coordinates for the  points  to stand   either side  in the road  according to direction to enter the vehicle - which presumably still applies to in street boarding

It may also  be useful to stand back and ask the so what why do you want to distinguish such stops - presumably the use case is to flag stops that are not suitable for disabled access. Doesn't an accessibility assessment do this more clearly?

PR #404, opened 2022-12-21
First proposal to deembed objects

skinkie:
From the experiences I have had over the past years with the actual use of the NeTEx datamodel, I have come to the realisation that one of the most problematic NeTEx points are generic object embeddings. An example of a generic embedding is a Quay which can currently only be placed under a StopPlace. This has the consequence that a StopPlace must always be defined in order to define a Quay.

Does NeTEx have a solution for it? Yes, one could use a GeneralFrame, which allows to embed any object in any order, but this also loses all the semantic context which are available between parent-child relations and frames.

What I would propose:
 1. allow any objecttype, where it would make sense and can be referenced, to be defined in a frame so a Quay and AccessSpace yes, but StopPointInJourneyPattern no.

Does this resolve everything? Certainly not. NoticeAssignment is a great example that can be applied to any object identifier, and would still require "magic" to correctly link. The other consequence is making inheritance and inference vague for many to one relationships. It is clear that a StopPlace/quays/Quay can infer content from its parent, but would the same go for StopPlace/quays/QuayRef?


Aurige:
Just as a side note: I fully agree that "deembeding" is sometimes useful, and that's one of the reasons why France choose the GeneralFrame as a base (and specialised it in the profile) since this already allows deembding (for example, with Quays, see the **Le Corbusier - SQYBUS-NeTEx-Profil Arret - External Quays.xml** file in the NeTEx SXD example hierarchy).
For some reason, this PR is not in our discussion list for the revision, shall I add it ?

ue71603:
yes, please.

ue71603:
@skinkie  what do we do with this PR?

skinkie:
@nick-knowles doesn't like it.

PR #409, opened 2023-01-31
Enhancement tmcr vehicle equipment profile member

nick-knowles:
Also includes example and constraints  for previous change to  vehicle types to add accepted driver permit.


skinkie (changes requested):
Again no.

ue71603:
@Aurige  @skinkie  do we need here a new clean one too?

ue71603:
integrated into: #612

PR #414, opened 2023-02-14
Fix vehicle equipment profile and ActualEquipmentProfile

nick-knowles:
Allow  multiple members  - Fixe needed for DEck plan change


skinkie:
Nope.

Aurige:
Better when targeting next
@skinkie : what is the best way to go in such situation where you have several commits in a raw, each one depending in the previous one ... this results in "non atomic" changes since the last PR collects commits (but need them all to be consistent).

skinkie:
I guess: push earlier?

Aurige:
yes, but not always possible/meaningful ... this seems to be a Github limitation to support dependent/stacked pull requests (https://stackoverflow.com/questions/26619478/are-dependent-pull-requests-in-github-possible )
In the case where we have one commit per PR, the easy way is probably to just focus on the latest commit

ue71603:
@trurlurl @skinkie  Also pls here, advise @trurlurl if changes are needed.

ue71603:
done in #615

PR #416, opened 2023-02-15
Single Files Validation issues

Aurige:
A few SXD files were not valid when used individually (netex_ifopt_equipmentAccess_support.xsd and netex_ifopt_equipmentSigns_support.xsd for example) 
This was solved by simply move PlaceEquipmentRef, PlaceEquipmentRefStructure and PlaceEquipmentIdType from netex_equipment_version.xsd to netex_equipment_support.xsd (which is  quite logical).

This is linked to the CR61 in the 2022-2023 NeTEx Revision


skinkie:
Is this for next or master?

Aurige:
I put it to Next in the context of the revision (I did it following the discussion of CR61)... but you are right, it may be on master (but that's not a big issue)

skinkie:
I would like to see bugs fixed on master and then propagated to next, that prevents 'problems'.

Aurige:
What is the "easy way" to change the target branch without collecting all previous merged Commits from next ?

skinkie:
Start with master for your bug fixes.
Apply master to next.
Continue with next.

ue71603:
fine with me, but we should do it in master, when it is a bug. Then I approvie it. Or is there a reason to do it here?

skinkie:
When used individually was something I missed. Maybe a stupid question: which files can be used individually?

Aurige:
replaced by https://github.com/NeTEx-CEN/NeTEx/pull/420

PR #417, opened 2023-02-15
Move from Travis to Github Actions

skinkie:
Once merged to master, this will lint all examples & xsd's (likely going to cause some merge issue) but I guess we should manage.


ue71603:
This changed almost every file. Which parts should I check?

skinkie:
@ue71603 just click on the first commit. 
https://github.com/NeTEx-CEN/NeTEx/pull/417/commits/8eced92187e44a9aaa5d480698f615a36b9df1ba

ue71603 (approved):

Aurige:
Ok for me (thanks for doing it).

As a side note, something I noted is that we don't have a strong validation of the XSD Files themselves: _xmllint_ checks the example against the XSD, which require a correct XSD, but this does not go in the details (that's why I have the "single File issue" PR, that was detected by XMLSpy but not xmllint). 
I tried with "`xmllint.exe --noout --schema http://www.w3.org/2001/XMLSchema.xsd  XSD-TO-VALIDATE`", it works but is missing some issues. I didn't find a comprehensive XSD validator outside of the Oxygen and SMLSpy tools. 
Nothing critical, but a possible enhancement.

Aurige (approved):

skinkie:
@Aurige @ue71603 @nick-knowles I'll try to do the big merge to next a bit later tonight.

PR #418, opened 2023-02-15
Fix add missing day type to duty

nick-knowles:
WD's  work on VDV duty identified that DAY TYPE was missing from DUTY - Added here


skinkie:
We continue in #684

PR #423, opened 2023-02-22
upload ci v2

skinkie:
To check if it now runs out of the box.


skinkie:
@Aurige @ue71603 it seems that this thing at least does the job when running with a local change.

ue71603 (approved):

Aurige (approved):

PR #447, opened 2023-06-14
Having multiple BookingNotes

ue71603:
In some cases multiple BookingNotes need to be send (espeicially when one uses text blocks).

### DIFF
--- a/xsd/netex_framework/netex_reusableComponents/netex_serviceRestrictions_version.xsd
+++ b/xsd/netex_framework/netex_reusableComponents/netex_serviceRestrictions_version.xsd
@@ -188,7 +188,7 @@ Rail transport, Roads and Road transport
 					<xsd:documentation>URL for booking. +V1.1</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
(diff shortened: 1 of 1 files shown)


skinkie (changes requested):
Don't like. This does not allow order, and it introduces an ungrouped set of elements. I would say make it BookingNotes (as choice).

Aurige:
I would also preferably go via a BookingNote**s** to group them ... or add a NoticeAssignment if you want an order (but you can already do it without the need of being explicit)

ue71603:
@Aurige @skinkie  We can remove BookingNote and make BookingArrangements a DataManagedobject, then Notice and NoticeAssignment are always available to BookingArrangements. I think this is better as Contact IS a DataManagedObject, but BookingArrangements is not. I redo the PR in this direction.

skinkie:
@ue71603 do you want to keep this since #525 is in my perspective allowing multiple bookingArrangements, each having a single note. So unless you want to go for _bookingNotes_, for something that is really beyond a multiline note. I don't like that you would have to go in the notice and noticeAssignment direction for simple cases.

ue71603:
Ok we close it.

PR #454, opened 2023-06-26
Introduce dataSourceRef checking, so we can see seriously broken examples and implementations

skinkie:
@Aurige my understanding is that @dataSourceRef + @version should point to DataSource @id + @version it becomes obviously once we enforce these constraints again a lot of examples just stop validating because they failed to implement the correct referening.

@ue71603 if we would agree to this principle, mentioned above having all EntitiesInVersion plus some 'rogue elements' pointing to the correct DataSource, we must update all the broken examples as well.


Aurige:
I'm note sure that I'm getting your point

Is it related to references like (from oe of the examples) :

`<AccessibilityAssessment version="001" created="2010-05-17T09:30:47Z" modification="new" dataSourceRef="tbd:accessibility_data" id="tbd:9100WIMBLDN@A1-EE1+5n6-acc">` (from one of the examples)

in that case the version is the one of the _AccessibilityAssessment_ not of the _dataSource_

Furthermore, in most case the DataSource will have one single version (NAPTAN is NAPTAN, SNCF is SNCF ...usually no sequence of version of them).

skinkie:
If a DataSource does not have the same version as the object it is on, that suggest that a single object can have concurrent version references. How would the _version_ of the DataSource be resolved (or provided) if there would be multiple instances of a DataSource with the same identifier?

Aurige:
Still lost : the data source is the data provider (DATA SOURCE identifies the system which has produced the data) ... the data provider is a stable entity. When UK is referencing the data source as being Naptan (as a system), are when it is Ile de France Mobilité (Paris' Authority, as their open data publication system) the data source version is not really meaningful. But the data source is possibly managing object's versions of course.
Do we put different concepts behind data source ?

skinkie:
If the DataSource is a stable entity it should **not** be versioned within NeTEx at the DataSource element level. There are multiple elements which do not have a version attribute and are truely unversioned.

xsd/NeTEx_publication.xsd
```diff
@@ -287,6 +287,11 @@ Correct COnstraints for PointOnRoute
...
+			<xsd:field xpath="@dataSourceRef"/>
+			<xsd:field xpath="@version"/>
```
skinkie, line 293: Should this be removed.

ue71603: line 293 removed.

ue71603:
Fixing examples for me.

ue71603:
I fixed it now so that it works for version, too. But will remove version now

ue71603 (approved):

trurlurl (approved):

xsd/NeTEx_publication.xsd
```diff
@@ -285,12 +285,17 @@ Correct COnstraints for PointOnRoute
...
+			<xsd:field xpath="@dataSourceRef"/>
+			<!-- <xsd:field xpath="@version"/>  We don't check the version for the dataSourceRef any longer. +v2.0 -->
 		</xsd:keyref>
 		<xsd:key name="DataSource_AnyVersionedKey">
 			<xsd:selector xpath=".//netex:DataSource"/>
 			<xsd:field xpath="@id"/>
-			<xsd:field xpath="@version"/>
+			<!-- <xsd:field xpath="@version"/> +v2.0 -->
```
skinkie, line 298: I think you should really remove it. Because from the perspective of the object refering to a DataSource it certainly will not be the same version.

xsd/NeTEx_publication.xsd
```diff
@@ -285,12 +285,15 @@ Correct COnstraints for PointOnRoute
 		<xsd:keyref name="DataSource_AnyKeyRef" refer="netex:DataSource_AnyVersionedKey">
 			<xsd:selector xpath=".//netex:DataSourceRef | .//netex:DefaultDataSourceRef"/>
 			<xsd:field xpath="@ref"/>
-			<xsd:field xpath="@version"/>
+			<!-- <xsd:field xpath="@version"/> +v2.0-->
```
skinkie, line 288: I am sorry I was stupid. But I think we are not there yet. I think for DataSourceRef we must have a version, but I should point to a DataSource key with id + version. Difficult.

ue71603: changed. better?

ue71603:
@skinkie @Aurige pls check

Issue #455, opened 2023-06-27
EPIP example wrong


skinkie:
We currently have an EPIP example in our repository that in my opinion has the wrong semantics going on for the version attribute.

https://github.com/NeTEx-CEN/NeTEx/blob/master/examples/standards/epip/epip_common_profile.xml#L49

Everything in that file refers to:
```
version="epip:1.0"
```

But reviewing the actual Version-element, it uses version="any" while version="epip:1.0" would be (in my opinion) the correct way.
```
						<Version id="epip:1.0" version="any">
							<Description>EPIP version 1.0</Description>
							<VersionType>baseline</VersionType>
						</Version>
```
Hence:
```
						<Version id="epip:1.0" version="epip:1.0">
							<Description>EPIP version 1.0</Description>
							<VersionType>baseline</VersionType>
						</Version>
```

What we are currently missing in our XSD is the check if a version actually exists.

Which then causes a major incompatibility with EPIP which explicitly state:

"In EPIP the Version element itself not populated; a simple version number in the version attribute of an EPIP element is used to indicate the version."

I think this deserves a change request, adding the Version element is _trivial_, especially if there is a single version anyways.

Aurige:
right, there is an issue here: the `<version ...>` is against the mention you are pointing and furthermore, looking deeper it looks that this version in the example is the version of the EPIP profile, and is use later on as a version of Data or Data set which is inconsistent.
And in any case, if the version is provided, I would expect something like `<Version id="epip" version="1.0">`
So I agree that this is a bug to correct

skinkie:
In that case I would expect version="1.0" on every element.

Aurige:
except if your providing system has no version management at all (which is often the case as explained in the same EPIP chapter), in that case no version is Ok (or "any" when you want/need to trig the constraint control)

skinkie:
Not ok for me, then it should provide a fixed value too.

ue71603:
@skinkie  only documentation to do? something else? we tighened the id requirement already.

skinkie:
Version-element deprecated. Can be closed. Only documentation to do.

Aurige:
Part 1 documentation was updated. No need in EPIP Document

PR #457, opened 2023-06-27
Fix version

skinkie:
This attempts to fix the issue #455.

Choices made:
 1. A Version element MUST exist in any file that uses versions.
 2. version="any" cannot be supported unless a Version with version="any" exists.
 3. This likely breaks a lot of examples too, so it is based on the other dataSourceRef branch.


ue71603 (approved):

skinkie:
@ue71603 still too many examples that should be fixed first.

Aurige:
see my comments on #455 
Also I don't agree with "_version="any" cannot be supported unless a Version with version="any" exists._" : "any" mean that you can refer to any version you got (having a version number or not). 
But I would agree that an object defined with an id and a version="any" does not really mean anything (but it's harmless)

skinkie:
It is not harmless. Especially because "any" has semantics. A default value specifically stating there is no version would work, since it hopefully wouldn't break constraint checking.

ue71603:
@skinkie  @Aurige  check the last commit. I worked on two examples and this adds layers that I am not sure, if they are good. Also "any" is NOT interpreted as it should. I had to add Datasource with id and version="78" to make it work.

Aurige:
It still looks weird to me (the version of a version, and with Id=version as attribute) it can be a convention, but does not really mean something and can be misunderstood... I personally would really prefer things like `<Version id="78">` for versions. Also defining a version with just an **_Id_** looks like a useless technical artefact to me: the version element was initially created to carry some semantics, not just to hold an Id.
This said, I agree that we need to clarify the semantic of "any"

ue71603:
@skinkie  will update this as we changed stuff already.

ue71603:
@skinkie  should I fix it before you start working? or will you do it anyhow?

PR #458, opened 2023-06-29
Harmonize with CEN master

erlendnils1:
CEN master changes since Mar 30, 2022 have been cherry picked where not in direct conflict with local changes  + some earlier formatting differences has been harmonized (in favor of CEN version)


erlendnils1:
Created in wrong repository

PR #462, opened 2023-07-10
Remove all abstract elements from constraint validation

skinkie:


Aurige:
This raises other questions like the fact that **ResponsibilityRole** is abstract, but not inherited by any other objects, and Roles are in all example I went through, use via **ResponsibilityRoleAssignment** that may use a **TypeOfResponsibilityRoleRef** but never a **ResponsibilityRoleRef** (since you can't instanciate a **ResponsibilityRole**).
so it looks that we can drop **ResponsibilityRole** (and associated **ResponsibilityRoleRef**)

Aurige:
I'd like to have @nick-knowles  's view on this

ue71603:
@nick-knowles  input needed here from you.

ue71603 (approved):

xsd/NeTEx_publication_timetable.xsd
```diff
@@ -1646,23 +1608,14 @@ Provides a general purose wrapper for NeTEx data content.</xsd:documentation>
...
 		<xsd:keyref name="Equipment_KeyRef" refer="netex:Equipment_AnyVersionedKey">
 			<xsd:selector xpath=".//netex:EquipmentRef"/>
 			<xsd:field xpath="@ref"/>
 			<xsd:field xpath="@version"/>
 		</xsd:keyref>
 		<xsd:key name="Equipment_AnyVersionedKey">
-			<xsd:selector xpath=".//netex:Equipment | .//netex:PlaceEquipment | .//netex:InstalledEquipment | .//netex:ActualVehicleEquipment | .//netex:TicketValidatorEquipment | .//netex:TicketingEquipment   | .//netex:SanitaryEquipment  | .//netex:PassengerSafetyEquipment | .//netex:RubbishDisposalEquipment    |.//netex:PassengerInformationEquipment   | .//netex:EntranceEquipment  | .//netex:LiftEquipment |  .//netex:EscalatorEquipment | .//netex:TravelatorEquipment | .//netex:StaircaseEquipment | .//netex:RampEquipment  | .//netex:RoughSurface  | .//netex:PlaceLighting|  .//netex:PlaceSign |  .//netex:PlaceSign | .//netex:HeadingSign  | .//netex:GeneralSign   |.//netex:WaitingRoomEquipment   | .//netex:ShelterEquipment  | .//netex:SeatingEquipment | .//netex:TrolleyStandEquipment      | .//netex:LuggageLockerEquipment  |  .//netex:CycleParkingEquipment    |.//netex:AccessVehicleEquipment    |.//netex:WheelchairVehicleEquipment"/>
```
trurlurl, line 1665: Is the removal of *LuggageLockerEquipment* intended and OK?

skinkie: Certainly, it is an abstract element.
./netex_part_1/part1_ifopt/netex_ifopt_equipmentWaiting_version.xsd:	<xsd:element name="LuggageLockerEquipment" abstract="true" substitutionGroup="SiteEquipment">

ue71603: @skinkie  closing the conversationthen?

trurlurl (approved):

Aurige:
I'm still puzzled here, and I think that the mistake is that the "ResponsibilityRole" element should not be abstract (there are plenty of refs to it, it's also as a first class element in Frame... and that's totally meaningless if it is abstract, and not inherited by any other element)

PR #471, opened 2023-07-26
Required IdType fixes

skinkie:
This pull request tries to address the missing required "id" attributes. This is selected by:
 - GeneralFrameMembers (they must have id, no question)
 - Keyrefs that are already modelled in the NeTEx_publication file

I may have added too many, which I certainly will review. For the following objects there are IdTypes defined, but there is no individual id attribute modelled, this should still be done.

```
CustomerService
Entrance
FareContractEntry_
GeneralFrameMember
OtherOrganisation
PassengerCarryingRequirementsView
PassingTimeView
Point
ServiceAccessRight_
SignEquipment
TravelSpecification_
TypeOfEntity
VehicleMeetingPlace_
Zone
```

PassingTimeView and PassengerCarryingRequirementsView are likely a bug in the schema that introduces them in the GeneralFrame. For PassingTimeView a DataManagedObjectView is first introduced, and never used after. For PassengerCarryingRequirementsView in my guess invalidly a DataManagedObject is used (see: #474)


skinkie:
Failing examples must be fixed. Already checked with the cardinality in the documentation.

xsd/netex_part_1/part1_ifopt/netex_ifopt_site_version.xsd
```diff
@@ -354,8 +354,43 @@ Rail transport, Roads and Road transport
 	</xsd:complexType>
 	<xsd:element name="Entrance" type="SiteEntrance_VersionStructure" abstract="false" substitutionGroup="SiteComponent">
 		<xsd:annotation>
-			<xsd:documentation>A physical entrance or exit to/from a SITE. May be a door, barrier, gate or other recognizable point of access.</xsd:documentation>
+			<xsd:documentation>A physical entrance or exit to/from a SITE. Maybe a door, barrier, gate or other recognizable point of access.</xsd:documentation>
 		</xsd:annotation>
+		<!-- TODO: Matthias
```
ue71603, line 359: Adding all this stuff is no good idea. It is already there in Entrance before. That is why the ambiguity occurs. 
An id is also there: 
image

Is the problem that it is not required?

skinkie: What we try to achieve here is that the attribute id gets overridden later with a specific ObjectIdType for that specific element. Theoretically we could add a regular expression then that would guard that the object is named like "codespace:ScheduledStopPoint:something". Not my intention now, but building the relationship between Ref and Id is.

xsd/netex_part_1/part1_ifopt/netex_ifopt_site_version.xsd
```diff
@@ -483,7 +483,7 @@ Rail transport, Roads and Road transport
 			</xsd:extension>
 		</xsd:complexContent>
 	</xsd:complexType>
-	<xsd:element name="VehicleEntrance" abstract="true" substitutionGroup="Entrance">
+	<xsd:element name="VehicleEntrance" abstract="true" substitutionGroup="SiteComponent">
```
skinkie, line 486: This kills the method resolution order (MRO) of the schema. Hence VehicleEntrance will be at the same level as Entrance, while what we want is having the inheritance.

ue71603: I think we have a problem already, because Entrance comes from SITE and VehicleEntrance probably should not be a SITE. Or do I read this wrong?

ue71603: It also might be one of the cases where we need an Entrance_ . Because having to restrictions in a row is not allowed in the inheritance. Should I try to model an Entrance_? According to one of the examples?

ue71603: Or we should do the inheritance through the VersionStructure elements. I think that only in the "leaves" the id stuff can be done once.

skinkie: I think this is a case where Entrance_ would make sense. In this situation I also want to figure out why (or not) Entrance_ may (or may not) be abstract.

ue71603:
fixed all assigned tasks

xsd/NX.xsd
```diff
@@ -37,7 +37,7 @@
 						<xsd:element name="points" minOccurs="0">
 							<xsd:complexType>
 								<xsd:sequence>
-									<xsd:element ref="Point" maxOccurs="unbounded"/>
+									<xsd:element ref="Point_" maxOccurs="unbounded"/>
```
skinkie, line 40: Are you sure about these changes? It feels like you are referencing to 'abstract' datatypes / 'containers'.

ue71603: I am an engineer (not true): It works... And it seems to me that they did it that way on other xxx_ elements.  What it does it references the "concrete" element that is used in the substitutionGroup. Otherwise it wouldn't work. But to be sure we would need a comment from Nick or Christophe.

skinkie: @nick-knowles @Aurige

Aurige:
I'm not sure that we can be so systematic ! A commented in other PR, I agree that any standalone object that can be referenced should be enforced to have an Id, but when it's an object that is embedded in the XML hierarchy and that will never be referenced, this constraint does not really makes sense (also it is harmless ... except that it brakes existing data).
For example, take the very first example you had to update : the `<LimitingRuleInContext>` is embedded, and giving it an Id is a constraint that  is useless.
I will add this as a discussion point fir next meeting.

xsd/NX.xsd
Aurige: This is a bit different from the topic of the PR ;-) but technically should be Ok: I think that this is replacing a ref to anything inheriting from Point to anything being of in the Point_ substitution group. I didn't check the possible semantic impact (yet).

skinkie:
Please notice that this did not break examples. If there are examples for those types, please add them.

skinkie:
@ue71603 in essence this is the problem with _order_ in https://github.com/skinkie/reference/issues/65 so if we on one hand have defined it as key element and thus compound key, and in the XML Schema it is optional. That is never going to work.

ue71603:
In my view objects have id.  One does not know if it will be referenced. They also may not be generated at the same time in the same process.

TuThoThai:
~~Group agreement:~~
~~- Ok for the fixes~~
~~- Can be merged after the examples are fixed by @skinkie~~

TuThoThai:
Edit as of 10 Oct. 2025

From plenary meeting (23 Sept. 2025): 
- Group agrees that `id` should always exist from v2.0 onwards.
- But cannot be targeted to `master`
- And has too many mixed commits
- Check consistency with constraints in netex_publication to avoid duplication of constraints

Actions:
- moved to milestone v2.1
- split the PR into several pieces

skinkie:
@TuThoThai 
I disagree with moving this to 2.1. This was agreed as late work for 2.0.

TuThoThai:
@skinkie, noted on your disagreement (disappointment).
We had to mitigate between realistic time we had to solve the conflicts, double check on possible constraints duplicates in NeTEx publication, and pressing requests to have a XSD ready for the CEN vote.

I suggest:
- we focus on the release of v2. 0
- we leave it targeting master for bug fixes of the future v2. 0
- if needed, we could move it to a "no milestone" and prioritise it for bug fixes

Would it be an acceptable route for you?

TuThoThai:
Release meeting as of 9 Oct 2025:
- Agreement to move it to v2.0 release
- @skinkie to rework it based on PR #962 
- @Aurige, @thbar and @TuThoThai to review
- Then, merge

Deadline: 14 Oct 2025

Aurige:
To complement yesterday's discussion, I wanted to be back on the fact that most IDs are already mandatory due to the constraints. For example, on ScehduleStopPoint we have (this is not uniqueness, that's a mandatory key)
```
		<xsd:key name="ScheduledStopPoint_AnyVersionedKey">
			<xsd:selector xpath=".//netex:ScheduledStopPoint | .//netex:FareScheduledStopPoint"/>
			<xsd:field xpath="@id"/>
			<xsd:field xpath="@version"/>
		</xsd:key>
```

Therefore if you try to write
```
				<ScheduledStopPoint>
					<Name>Poste, St Jean</Name>
					<Location>
						<Longitude>-0.2071397147</Longitude>
						<Latitude>51.4217482061</Latitude>
					</Location>
					<VehicleModes>bus</VehicleModes>
				</ScheduledStopPoint>
```
You will get an error ... this on in XMLSpy
image

so the **use="required"** added on IDs, is really doubling this constraint, furthermore is is not comprehensive since it it is not including the **version** part of the constraint. So why not just adding the possibly missing constraints ?

In any case, this PR does not harm and I won't be against it, but I'm still not sure that it is the best way to go.

skinkie:
Code generators (Java, Python, etc.) won't know anything about xsd:keys. They only consider the attribute.
The identity key contraints fall apart if attributes are missing. So if field @id or @version is missing: it won't even report an error.

xsd/netex_framework/netex_genericFramework/netex_accessibility/netex_acsb_accessibility.xsd
```diff
@@ -90,7 +90,7 @@ Rail transport, Roads and Road transport
 							<xsd:group ref="SensoryLimitationGroup"/>
 						</xsd:sequence>
 					</xsd:sequence>
-					<xsd:attribute name="id" type="LimitationIdType"/>
+					<xsd:attribute name="id" type="LimitationIdType" use="optional"/>
```
trurlurl, line 93: This is one of very few cases of optional uses. I guess this is as intended and it just is not clear to me why.

TuThoThai:
Unless mistaken, this one can be closed as it was addressed by the combo of #970 and #972 

Could you confirm the status, @skinkie?

skinkie:
@TuThoThai I guess no champagne on Friday?

TuThoThai:
As of 13 March 2026: please @skinkie, make the call if the PR is still needed or not

PR #14, opened 2018-04-03
Rename all xsd files (and references to them) to not contain version

csolem:
See #13

Change was done using two bash one-liners:
```
find . -iname "*.xsd" | while read file; do newName=$(echo $file | sed 's/-v1\.[0-9]//g');  if [ "$file" != "$newName" ]; then echo "$file -> $newName"; mv $file $newName; fi; done;
```

```
find . -iname "*.xsd" | while read file; do sed -i 's/-v1\.[0-9]//g' $file; done;
```

Changes have been confirmed to work with xmllint and a valid netex xml file:
```
 xmllint --noout --schema xsd/NeTEx_publication.xsd some_netex_file.xml
```

Verification of compatibility with other currently unmerged pull requests:
I have tested that it is possible to merge this code into the branch containing #9

PR #477, opened 2023-07-28
Retype all order elements to xsd:positiveInteger and fix the associated examples

### LINKED ISSUE #475: Investigate order attribute xsd:integer vs xsd:positiveInteger

skinkie:
This is a breaking change (with respect to the documentation) which mixes xsd:positiveInteger with xsd:integer.

The main difference is: positiveInteger starts at 1.

Please carefully consider the changes that have a mention on "Relative order". No examples exists with a negative relative order, but the design might allow them?

Fix #475

(apply after the required Id pull request)


trurlurl (approved):
Looks good to me.

trurlurl:
@skinkie No, I cannot be sure about negative order numbers, I'm not sufficiently familiar with the structures in question.

ue71603:
I think we moved the elevator order into an element. So it means that order is arbitrary.  And from N to Q everything can be ordered the same way ;-)

ue71603 (approved):

skinkie:
Show me the breaking data.

Aurige:
> Show me the breaking data.

The examples you had to correct ?

skinkie:
> The examples you had to correct ?

The PassengerStopAssignment order should always have started at 1?

Aurige:
Will be discussed in meeting #14 (October 4th)

ue71603:
> Will be discussed in meeting #14 (October 4th)

Does this mean you are ok with it?

Aurige:
> > Will be discussed in meeting #14 (October 4th)
> 
> Does this mean you are ok with it?

My mind is not fully made up about it (not against, but don't like its drawbacks)... and anyway whatever I prefer, we need the group validation to accept the drawbacks

ue71603:
@Joostb61  RelativeLevelOrder is not affected. Only the order attributes are concerned.

Aurige:
The group decision was : The PR is accepted but needs to be separated from the IdType fix on which it is stacked
_Note: may generate some backward compatibility issues_

ue71603:
@skinkie  who will remove  the IdType part?  Should I rebase it?. Otherwise it was accepted by enough reviewers. Or do you want explicitly have acceptance by @Aurige

skinkie:
If you only go for the changing or order element, I would cherry pick the specific commits.

ue71603:
use the last four or so comments. and redo...

Aurige:
How to express that order (mandatory) is not useful in some cases ? Make it non mandatory ? use value "1" for all orders ? 
Value 1 is the group's preferred choice ... to be described in the document (only when the order is meaningless)

trurlurl:
Removing "documentation update needed" since closed without merging.

PR #411, opened 2023-02-06
Extended-DatedVehicleJourney+Replace-PR368

Aurige:
Initial PR368 from Kristian Syversen (Kirstian has moved and the PR was not editable ... so it was easier to create a new one). 

This is the exact copy of PR368, but targeting 'next' and not master branch 
It complements it with deprecated DriverRef (DatedJourney are expected to be referenced by driver's DUTY, not the way arround). 

This is conforming the decision from meeting #3 of the NeTEx revision 2022/2023.

Issue #13, opened 2018-03-21
Remove version suffixes from all files


skinkie:

skinkie:
Thanks @csolem

PR #518, opened 2023-10-06
Extended dated vehicle journey+replace pr368 and pr411

Aurige:
Update of the [Pr411 ](https://github.com/NeTEx-CEN/NeTEx/pull/411) (that was merged and rolled back due to issues with examples uising UIC periods).
This new version integrates the decisions from the Meeting #13 on October 5th:
- Update of examples to use OperatingDayRef in dated jouneys (keep the old UIC periods commented with an explicit mention of the deprecation).
- UicOperatingPeriod deletion is now replaced by a simple deprecation (clearly stated in comments) for backward compatibility reasons.


skinkie:
@Aurige you make a reference to an OperatingDay, is this defined in the file?

Aurige:
> @Aurige you make a reference to an OperatingDay, is this defined in the file?

no, it is most likely that OpertatingDays can be shared across multiple such files and therefore exchanged separately in a calendar frame.

skinkie:
Lets not go into the direction that we share examples that do not validate on their own.

Aurige:
> Lets not go into the direction that we share examples that do not validate on their own.

I checked validation with XML Spy (and external Ref are OK ...)

examples/standards/era_uic/Netex_era_uic_joiningsplitting.xml
```diff
@@ -515,11 +515,13 @@ This example shows the usage of coupled journeys, footnodes and two directions f
 									<LuggageCarriageFacilityList>cyclesAllowedWithReservation</LuggageCarriageFacilityList>
 								</ServiceFacilitySet>
 							</facilities>
+							<OperatingDayRef ref="tap:OperatingDay:2009-03-21"/>
```
skinkie, line 518: This is not an external ref, and is missing a version.

examples/standards/era_uic/Netex_era_uic_joiningsplitting.xml
```diff
@@ -515,7 +515,7 @@ This example shows the usage of coupled journeys, footnodes and two directions f
 									<LuggageCarriageFacilityList>cyclesAllowedWithReservation</LuggageCarriageFacilityList>
 								</ServiceFacilitySet>
 							</facilities>
-							<OperatingDayRef ref="tap:OperatingDay:2009-03-21"/>
+							<OperatingDayRef ref="tap:OperatingDay:2009-03-21" versionRef="1"/> <!-- reference to an  OperationalDay defined in another dataset -->
```
skinkie, line 518: Just no. There is no reason **not** to include it in this example.

EU-Agency-for-Railways (changes requested):
I reviewed the changes on the examples and the restriction to one operational day makes them useless for the purpose to explain the usage of NeTEx to replace the existing EDIFACT messages for the timetable data exchange for rail. Rail timetables are usually not planned for a specific operation day. So there is always a validity period plus a bitmask explaining the operating days. I would suggest to redraft the examples with a more suitable element (e.g. ServiceJourney).

Aurige:
thanks @EU-Agency-for-Railways 
I agreed that the best way to go is to rework these examples and built new cleaner ones using ServiceJourney
This raised a few questions:

- What do we do with the current examples ? do we just temporary drop them or deprecate them, waiting for the new version ?
- Who can contribute to building new examples ? It's quite  lot of examples to update (15 examples, some being quite long)... expect probably a few days to fully redesign and rebuilt them.

Also I would like to separate the XSD from the XML example updates, so it this is Ok for you, I'm going to remove the examples from this PR and create a new separate one for their update.

Aurige:
@skinkie I reverted all changes to examples but had issues with other commits that were automatically brought back by Github and that were also modifying and breaking this PR... I had to revert it, but now it's a total mess ... I'm not sure what happened, but I'm afraid that I need to recreate this PR once more !!

skinkie:
@Aurige if you have a lot of free time, I have more interesting things to discuss ;-)

Aurige:
replaced by https://github.com/NeTEx-CEN/NeTEx/pull/520

PR #520, opened 2023-10-13
Rail dated journey fix without example updates

Aurige:
Update of the Pr411 (that was merged and rolled back due to issues with examples uising UIC periods) and PR518 (broken for whatever github reason)

This version integrates the decisions from the Meeting # 13 and # 14:

Update of examples to use OperatingDayRef in dated jouneys (keep the old UIC periods commented with an explicit mention of the deprecation).
UicOperatingPeriod deletion is now replaced by a simple deprecation (clearly stated in comments) for backward compatibility reasons.


skinkie (approved):

ue71603 (approved):

Aurige:
Part 2 was updated

PR #529, opened 2023-10-19
Fix missing attributes Alternative Text

skinkie:


ue71603 (approved):

Aurige:
What is the difference between the **lang** we already have and the **useForLanguage** that is added here ?

image

I would see this PR as en enhancement more than a bug (and therefore target Next)

skinkie:
> What is the difference between the **lang** we already have and the **useForLanguage** that is added here ?

No clue, that attribute is used in the _restriction base_.

Aurige:
Ok, sorry I missed your point .. I agree that this correction looks necessary to get the proper restriction... also it currently works without it (examples are using **order** and **useForLanguage** )

skinkie:
> Ok, sorry I missed your point .. I agree that this correction looks necessary to get the proper restriction... also it currently works without it (examples are using **order** and **useForLanguage** )

That is an upstream bug.

skinkie:
Can be closed, not a bug.

PR #534, opened 2023-10-31
Fix #533

### LINKED ISSUE #533: Missing LowerHandrailHeight in RampEquipment
Hello,
is there a reason why we can find the XSD-Element LowerHandrailHeight in the StairGroup but not in the RampEquipmentGroup? In our opinion we should also describe the height of the lower handrail in RampEquipment. What do you think?

Here the example of LowerHandrailHeight in the StairGroup.
https://github.com/NeTEx-CEN/NeTEx/blob/7a1af73cf6060c813c0be272fd77a2eaabf821d7/xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_version.xsd#L382

Kind regards,
David

skinkie:
Reorder elements so they have the same order as the other groups. Add the LowerHandrailHeight as requested in #533.

### DIFF
--- a/xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_version.xsd
+++ b/xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_version.xsd
@@ -1036,7 +1036,7 @@ Rail transport, Roads and Road transport
 			</xsd:element>
 			<xsd:element name="LowerHandrailHeight" type="LengthType" minOccurs="0">
 				<xsd:annotation>
-					<xsd:documentation>Height of any additional lower handrail from step.</xsd:documentation>
+					<xsd:documentation>Height of any additional lower handrail.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
 			<xsd:element name="TactileGuidanceStrips" type="xsd:boolean" minOccurs="0">
--- a/xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_version.xsd
+++ b/xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_version.xsd
@@ -1024,14 +1024,19 @@ Rail transport, Roads and Road transport
 					<xsd:documentation>Whether ramp is on a pedestal.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
+			<xsd:element name="HandrailType" type="HandrailEnumeration" minOccurs="0">
+				<xsd:annotation>
+					<xsd:documentation>Type of handrail.</xsd:documentation>
+				</xsd:annotation>
+			</xsd:element>
 			<xsd:element name="HandrailHeight" type="LengthType" minOccurs="0">
 				<xsd:annotation>
 					<xsd:documentation>Height of handrail.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
-			<xsd:element name="HandrailType" type="HandrailEnumeration" minOccurs="0">
+			<xsd:element name="LowerHandrailHeight" type="LengthType" minOccurs="0">
 				<xsd:annotation>
-					<xsd:documentation>Type of handrail.</xsd:documentation>
+					<xsd:documentation>Height of any additional lower handrail from step.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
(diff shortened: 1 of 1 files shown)


xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_version.xsd
```diff
@@ -1024,14 +1024,19 @@ Rail transport, Roads and Road transport
...
 			<xsd:element name="HandrailHeight" type="LengthType" minOccurs="0">
 				<xsd:annotation>
 					<xsd:documentation>Height of handrail.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
-			<xsd:element name="HandrailType" type="HandrailEnumeration" minOccurs="0">
+			<xsd:element name="LowerHandrailHeight" type="LengthType" minOccurs="0">
 				<xsd:annotation>
-					<xsd:documentation>Type of handrail.</xsd:documentation>
+					<xsd:documentation>Height of any additional lower handrail from step.</xsd:documentation>
```
DWillDB, line 1039: I would suggest to delete "from step". Ramps have no steps.
```suggestion
					<xsd:documentation>Height of any additional lower handrail.</xsd:documentation>
```

DWillDB (approved):

trurlurl (approved):

ue71603 (approved):

Aurige:
I'm not that happy with reordering ... this one is, for example, generation a compatibility issue with the Accessibility tool we have developed in France... therefore, as French representative, I can't approve it without our mirror group validation (which I most probably won't get on that point).

skinkie:
@Aurige then you should have reported this inconsistency earlier, like the nice people of Germania do.

Aurige:
ha ha ... some even smarter may have raised some comments at CIB time, but we can't replay that game unfortunately.
More seriously, we must refrain from merging PR that are breaking compatibility before getting the validation from the NeTEx working group, the small team working on Github is unfortunately not representative enough for this kind of decision.

skinkie:
EPIAP has not even been voted for, so no. If people like from DB come with serious issues in the technical artifact, we fix the technical artifact. If some organisation started to work with a branch that has not even been merged into master, the only good action would be to update a line of code in their software. Or, better: inform them that they should just regenerate their automatically generated code.

Aurige:
I'm not talking about the EPIAP branch, but about the master ... and I guess that EPIAP will be merged to master ... so no, you can't decide to break compatibility on your own.

skinkie:
This pull request is on the EPIAP-branch and I have not decided on my own. In addition to that, there is no example that breaks. Please ask your French friends to contribute to this repository by at least providing the _output_ of their tool for our example folder.

Aurige:
The French profile for accessibility is here https://normes.transport.data.gouv.fr/normes/netex/accessibilit%C3%A9/ and here https://github.com/etalab/transport-normes and here http://www.normes-donnees-tc.org/profils/ for years and a lot of people were involved including several disabled person associations. Multiple ongoing work is now based on this as you can see here https://www.ecologie.gouv.fr/donnees-daccessibilite. And all this was started long before EPIAP.

But what I'm saying is that compatibility breaks cannot be accepted without a group decision ! You won't change my mind on this point.

skinkie:
The only way to assure that no downstream works breaks by pull requests, and the ability to inform them, is to have their output. That is what a national representative should guarantee. That obviously means that if downstream profiles are unable to be validated due to different interpretation of referential constraints, or private additions they would not meet our standard.

You reference a lot of paper trails, but no actual XML-document that would break because of this change.

Aurige:
Paper comes first ... if we change the rules while people are applying them, we are fired
Also there is no data provisioning system that use every little feature of a profile, so examples will never cover everything... but I fully agree that we need to have them.
And here, you don't need an example, I can tell you that this breaks the compatibility with the French profile, it is just a fact, whether you and I like it or not.

skinkie:
I think you will agree that with the merge of EPIAP (regardless of this change) the 'French profile' would break. But I am also happy that we agree that we need actual examples.

Aurige:
Yes sure ... I'm not against any single break, but we need a group discussion and a good reason for it. It will be at the agenda of next meeting.

Aurige:
Ok for the group... no more action requested

PR #552, opened 2023-11-08
Add external ref to StopPlace and Quay (Fix #451)

### LINKED ISSUE #451: Data Element for undecorated object id
In some EU countries (Germany, Austria, Switzerland, UK to name a few) national standards for global identifiers of stops and other important objects existed before the arrival of the EPIP standard. These existing identifiers are used to build the EPIP conformant NeTEx IDs:
For instance Salzburg Hauptbahnhof has national ID at:45:50002, where at is for Austria, 45 is for Oberösterreich (federal state) and 50002 is the local stop number within Oberösterreich. In NeTEx, Salzburg Hauptbahnhof looks like this

image

The national ID goes into the EPIP technical identifier, but it needs to be transformed to make it EPIP compliant - the ":" need to be replaced by "-" and the at: has to be removed and moved to the front of the EPIP id. Also "svv" as responsible authority is added.

However, all technical systems use the national ID in its original form and they would have to somehow interpret the EPIP id to determine the national ID when importing data. As a workaround, the original national ID is now stored in keyList/keyValue/Key.
The German VDV workgroup proposes to add a dedicated field "technical identifier" or "global id" containing the unmodified id to all NeTEx objects (or at least the relevant ones).

duexw:
This is the agreed fix for issue #451 .
New Elements ExternalStopPlaceRef and ExternalQuayRef are added to StopPlace and Quay. 
They can be used for storing national stop and quay identifiers in the case where these do not conform to the ID format described in EPIP .

### DIFF
--- a/xsd/netex_framework/netex_genericFramework/netex_assignment_version.xsd
+++ b/xsd/netex_framework/netex_genericFramework/netex_assignment_version.xsd
@@ -60,7 +60,7 @@ Rail transport, Roads and Road transport
 	<!-- ======================================================================= -->
 	<xsd:element name="Assignment_" type="DataManagedObjectStructure" abstract="true" substitutionGroup="DataManagedObject">
 		<xsd:annotation>
-			<xsd:documentation>Dummy Abstract Assignment. An Assignment assigns a property to an other element. It has a name and an order.</xsd:documentation>
+			<xsd:documentation>Dummy Abstract Assignment. An Assignment assigns a property to another element. It has a name and an order.</xsd:documentation>
 		</xsd:annotation>
 	</xsd:element>
 	<xsd:complexType name="Assignment_VersionStructure_" abstract="true">
--- a/xsd/netex_framework/netex_genericFramework/netex_assignment_version.xsd
+++ b/xsd/netex_framework/netex_genericFramework/netex_assignment_version.xsd
@@ -82,7 +82,7 @@ Rail transport, Roads and Road transport
 	</xsd:complexType>
 	<xsd:group name="AssignmentGroup">
 		<xsd:annotation>
-			<xsd:documentation>Elements for an ASSIGNMENT. An ASSIGNMENT assigns a property to an another element. It has a name and an order.</xsd:documentation>
+			<xsd:documentation>Elements for an ASSIGNMENT. An ASSIGNMENT assigns a property to another element. It has a name and an order.</xsd:documentation>
 		</xsd:annotation>
 		<xsd:sequence>
 			<xsd:element name="Name" type="MultilingualString" minOccurs="0">
--- a/xsd/netex_framework/netex_genericFramework/netex_assignment_version.xsd
+++ b/xsd/netex_framework/netex_genericFramework/netex_assignment_version.xsd
@@ -99,7 +99,7 @@ Rail transport, Roads and Road transport
 	</xsd:group>
 	<xsd:element name="Assignment" abstract="true" substitutionGroup="Assignment_">
 		<xsd:annotation>
-			<xsd:documentation>A set of properties to be applied to an another element. It has a name and an order.</xsd:documentation>
+			<xsd:documentation>A set of properties to be applied to another element. It has a name and an order.</xsd:documentation>
 		</xsd:annotation>
 		<xsd:complexType>
 			<xsd:complexContent>
--- a/xsd/netex_framework/netex_genericFramework/netex_assignment_version.xsd
+++ b/xsd/netex_framework/netex_genericFramework/netex_assignment_version.xsd
@@ -60,12 +60,12 @@ Rail transport, Roads and Road transport
 	<!-- ======================================================================= -->
 	<xsd:element name="Assignment_" type="DataManagedObjectStructure" abstract="true" substitutionGroup="DataManagedObject">
 		<xsd:annotation>
-			<xsd:documentation>Dummy Abstract Assignment An Assignment assigns a property to an assignment element. It has a name and n order.</xsd:documentation>
+			<xsd:documentation>Dummy Abstract Assignment. An Assignment assigns a property to an other element. It has a name and an order.</xsd:documentation>
 		</xsd:annotation>
 	</xsd:element>
 	<xsd:complexType name="Assignment_VersionStructure_" abstract="true">
 		<xsd:annotation>
-			<xsd:documentation>Type for  ASSIGNMENT.</xsd:documentation>
+			<xsd:documentation>Type for ASSIGNMENT.</xsd:documentation>
 		</xsd:annotation>
 		<xsd:complexContent>
 			<xsd:extension base="DataManagedObjectStructure">
@@ -74,7 +74,7 @@ Rail transport, Roads and Road transport
 				</xsd:sequence>
 				<xsd:attribute name="order" type="xsd:integer">
 					<xsd:annotation>
-						<xsd:documentation>Order in which to show  ASSIGNMENT,</xsd:documentation>
+						<xsd:documentation>Order in which to show an ASSIGNMENT.</xsd:documentation>
 					</xsd:annotation>
 				</xsd:attribute>
 			</xsd:extension>
@@ -82,7 +82,7 @@ Rail transport, Roads and Road transport
 	</xsd:complexType>
 	<xsd:group name="AssignmentGroup">
 		<xsd:annotation>
-			<xsd:documentation>Elements for an ASSIGNMENT. An Assignment assignes a property to an aelent element. It has a name and n order.</xsd:documentation>
+			<xsd:documentation>Elements for an ASSIGNMENT. An ASSIGNMENT assigns a property to an another element. It has a name and an order.</xsd:documentation>
 		</xsd:annotation>
 		<xsd:sequence>
 			<xsd:element name="Name" type="MultilingualString" minOccurs="0">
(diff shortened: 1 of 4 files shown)


skinkie:
@duexw you are now targeting master, but this change is actually for the future schema. That having said, why not use our new privateCodes structure, which may include the "AVMS" as a type?

duexw:
We would rather not like to use private code, because the intended use is by definition not private. We have some global identifiers in the German VDV standards, which are used to identify the objects across system borders.

We would like to express them in NeTEx as follows

- DLID/DTID (VDV 433): global Line ID: -> ExternalLineRef
- DFID (VDV 433): global Journey ID: -> ExternalVehicleJourneyRef
- DHID (VDV 432): global stop point id -> ExternalStopPointRef
- DHID (VDV 432): global quay  -> ExternalQuayRef
- DHID (VDV 432): global stop place ref  -> ExternalStopPlaceRef

Aurige:
It's Ok for me but I'd rather target it to next

skinkie:
I would like to have a discussion about this. In my perspective the whole 'external' thing is something that doesn't make sense to me. If it is _external_ it is by definition _private_ otherwise that external system would have used the NeTEx id in the first place.

duexw:
OK, I changed the branch from master to next. 
@skinkie : I do not remember why we used the prefix "External" for ExternalLineRef and ExternalVehicleJourneyRef back in the day when these elements were created - but I thought it better to stay in line with the two new ones. And I agree that ideally systems should use the id value directly. However, there are cases where this is not possible. For instance, in VDV 433 the DFID = ExternalVehicleJourneyRef needs to be unique only for one operating day, whereas the Journey id would have to be unique across all journeys and operating days.

xsd/netex_framework/netex_reusableComponents/netex_mode_support.xsd
```diff
@@ -68,19 +68,23 @@ Rail transport, Roads and Road transport
...
 		</xsd:annotation>
 		<xsd:restriction base="xsd:string">
 			<xsd:enumeration value="foot"/>
 			<xsd:enumeration value="bicycle"/>
 			<xsd:enumeration value="boat"/>
 			<xsd:enumeration value="car"/>
-			<xsd:enumeration value="taxi"/>
+			<xsd:enumeration value="taxi">
+				<xsd:annotation>
+					<xsd:documentation>Taxi can now be modeled as VEHICLE MODE as well.</xsd:documentation>
```
trurlurl, line 85:
```suggestion
					<xsd:documentation>Taxi can now be modelled as VEHICLE MODE as well.</xsd:documentation>
```

xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_support.xsd
```diff
@@ -348,6 +348,11 @@ Rail transport, Roads and Road transport
 					<xsd:documentation>Short public code for passengers to use when uniquely identifying the stop by SMS and other self-service channels.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
+			<xsd:element name="ExternalStopPlaceRef" type="ExternalObjectRefStructure" minOccurs="0">
+				<xsd:annotation>
+					<xsd:documentation>An alternative code that uniquely identifies the STOP PLACE. Specifically for use in AVMS systems that require an alias, if. For VDV compatibility.</xsd:documentation>
```
trurlurl, line 353:
```suggestion
					<xsd:documentation>An alternative code that uniquely identifies the STOP PLACE. Specifically for use in AVMS systems that require an alias. For VDV compatibility.</xsd:documentation>
```

xsd/netex_part_1/part1_ifopt/netex_ifopt_stopPlace_version.xsd
```diff
@@ -573,6 +573,11 @@ contained within its parent QUAY.</xsd:documentation>
 					<xsd:documentation>A 20 bit number used for wireless cleardown of stop displays by some AVL systems.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
+			<xsd:element name="ExternalQuayRef" type="ExternalObjectRefStructure" minOccurs="0">
+				<xsd:annotation>
+					<xsd:documentation>An alternative code that uniquely identifies the QUAY. Specifically for use in AVMS and journey planning systems that require an alias, if. For VDV compatibility.</xsd:documentation>
```
trurlurl, line 578:
```suggestion
					<xsd:documentation>An alternative code that uniquely identifies the QUAY. Specifically for use in AVMS and journey planning systems that require an alias. For VDV compatibility.</xsd:documentation>
```

trurlurl (changes requested):

ue71603 (approved):

xsd/netex_framework/netex_reusableComponents/netex_mode_support.xsd
```diff
@@ -273,7 +273,7 @@ Rail transport, Roads and Road transport
 			</xsd:enumeration>
 			<xsd:enumeration value="funicular">
 				<xsd:annotation>
-					<xsd:documentation>cable railway on steep slope using two counterbalanced carriages.</xsd:documentation>
+					<xsd:documentation>cable railway on steep slope, typcially using two counterbalanced carriages.</xsd:documentation>
```
skinkie, line 276:
```suggestion
					<xsd:documentation>cable railway on steep slope, typically using two counterbalanced carriages.</xsd:documentation>
```

xsd/netex_framework/netex_genericFramework/netex_assignment_version.xsd
```diff
@@ -74,15 +74,15 @@ Rail transport, Roads and Road transport
...
+						<xsd:documentation>Order in which to show an ASSIGNMENT.</xsd:documentation>
 					</xsd:annotation>
 				</xsd:attribute>
 			</xsd:extension>
 		</xsd:complexContent>
 	</xsd:complexType>
 	<xsd:group name="AssignmentGroup">
 		<xsd:annotation>
-			<xsd:documentation>Elements for an ASSIGNMENT. An Assignment assignes a property to an aelent element. It has a name and n order.</xsd:documentation>
+			<xsd:documentation>Elements for an ASSIGNMENT. An ASSIGNMENT assigns a property to an another element. It has a name and an order.</xsd:documentation>
```
skinkie, line 85:
```suggestion
			<xsd:documentation>Elements for an ASSIGNMENT. An ASSIGNMENT assigns a property to another element. It has a name and an order.</xsd:documentation>
```

xsd/netex_framework/netex_genericFramework/netex_assignment_version.xsd
```diff
@@ -60,12 +60,12 @@ Rail transport, Roads and Road transport
 	<!-- ======================================================================= -->
 	<xsd:element name="Assignment_" type="DataManagedObjectStructure" abstract="true" substitutionGroup="DataManagedObject">
 		<xsd:annotation>
-			<xsd:documentation>Dummy Abstract Assignment An Assignment assigns a property to an assignment element. It has a name and n order.</xsd:documentation>
+			<xsd:documentation>Dummy Abstract Assignment. An Assignment assigns a property to an other element. It has a name and an order.</xsd:documentation>
```
skinkie, line 63:
```suggestion
			<xsd:documentation>Dummy Abstract Assignment. An Assignment assigns a property to another element. It has a name and an order.</xsd:documentation>
```

Aurige (approved):

skinkie:
I would like to discuss this thing the the merge feast.

skinkie:
@ue71603 it should still work. Especially for everyone that has upgraded their libxml2 stuff :-)

ue71603:
Nick: Must be moved to SiteComponent
Stefan dK: Why having an External anyway. When we have KeyList. Argument understandable. I would like to remove it from the other object.
Stefan J.: Same.
Stefan dK: PrivateCode also extended.
Wilfried: Then we have a double concept already. In den VDV 462 group it was discussed. They wanted it that way. Dedicated element for this. KeyList is too unspecific.  These id are prime citizens. 
Matthias: SNCF does all id in KeyList. Laurent in Switzerland wants to do it this way as well  Key:"SLOID" Value:"ch:1:StopPlace:8507000"
Nick: I would see it.
Stefan dK: But this is the definition of PrivateCode.
Nick: We should clarify the difference.

ue71603:
Christophe: IdF has these things.
Wilfried: PrivateCode is very local. The national identifier are nationwide. They are cross-system.
Wilfried: A main problem was the ":"
Wilfried: Stop, Quay etc. it is ok. In VehicleJourney we have the external reference.

Wilfried: We can close it, when the EPIP close.  Removal of the other entries. Stefan creates a PR to remove it. 

=> EPIP will be done by Christophe.
=> PR for removal Stefan

Aurige:
There is an overlap between External Refs, PrivateCode and Key lists (more generic extension mechanism)  ... 
The EPIP doc needs to be updated accordingly on the Id codification
Group decision: stick on existing and close

ue71603:
Wilfried: We have the German national identifier. Is valid on one day.

skinkie:
This one should be closed.

nick-knowles:
Note that  tif one uses just  CODESPACEs, then  there  probably isnt a need for a private /external code (unless there is a different legacy code being exchanged)

Thus rather than
    <StopPlace id="ch:1:StopPlace:8507000">
       <PrivateCode>8507000/>
      instead just
    <StopPlace ="sncf:8507000">

where Codespace "sncf"  is understood to mean the lexical scope for sncf stops

ue71603:
> Note that tif one uses just CODESPACEs, then there probably isnt a need for a private /external code (unless there is a different legacy code being exchanged)
> 
> Thus rather than 8507000/> instead just <StopPlace ="sncf:8507000">
> 
> where Codespace "sncf" is understood to mean the lexical scope for sncf stops

In theory yes. In practice often the id are generated by an aggregation enitity and sometimes are only stable during an export. so having an additional id is certainly not a bad idea.

trurlurl:
Removing "documentation needed" label since closed withoud merging.

PR #548, opened 2023-11-04
Add test crossing equipment example

skinkie:
Fix #361

PR #559, opened 2023-11-24
Relative level order optional

ue71603:
fixes the point that Christophe made in #548


skinkie:
You want to go on next or epiap?

ue71603:
This one to next. as it vexed Christophe. In EPIAP it is right already

Aurige (approved):

Issue #452, opened 2023-06-20
Align NeTEx with OJP documentation for FareQuota


skinkie:
https://github.com/VDVde/OJP/pull/371/files

ue71603:
addressed in #590

PR #590, opened 2023-12-12
add FareQuota

ue71603:
fixes https://github.com/NeTEx-CEN/NeTEx/issues/452

### DIFF
--- a/xsd/netex_part_3/part3_fares/netex_fareProduct_version.xsd
+++ b/xsd/netex_part_3/part3_fares/netex_fareProduct_version.xsd
@@ -1137,7 +1137,7 @@ Rail transport, Roads and Road transport
 			<xsd:documentation>Pricing Elements for a FARE PRODUCT PRICE.</xsd:documentation>
 		</xsd:annotation>
 		<xsd:sequence>
-			<xsd:element name="FareQuota" type="xsd:integer" minOccurs="0">
+			<xsd:element name="FareQuota" type="xsd:nonNegativeInteger" minOccurs="0">
 				<xsd:annotation>
 					<xsd:documentation>Offered tickets in this FARE PRODUCT. When absent, the number of remaining tickets is unknown or it is not limited. In most cases if the FareQuota is zero then the FareProduct is sold out.</xsd:documentation>
 				</xsd:annotation>


trurlurl (approved):

xsd/netex_part_3/part3_fares/netex_fareProduct_version.xsd
```diff
@@ -1137,6 +1137,11 @@ Rail transport, Roads and Road transport
 			<xsd:documentation>Pricing Elements for a FARE PRODUCT PRICE.</xsd:documentation>
 		</xsd:annotation>
 		<xsd:sequence>
+			<xsd:element name="FareQuota" type="xsd:integer" minOccurs="0">
```
skinkie, line 1140:
```suggestion
			<xsd:element name="FareQuota" type="xsd:nonNegativeInteger" minOccurs="0">
```

Aurige:
We need to discuss this: is the FareProductPrice a really good place for a quota ? Also, what does the quota applies to (the full FARE PRODUCT, a specific JOURNEY to which the product gives access to, etc.).
Up to now we have the FareQuotaFactor and you can assign it via a ValidityPartameterAssignment to whatever it concerns.

ue71603:
@Aurige No worries. We can also NOT do it. We had it done in OJP. In OJPFareResponse a quota makes sense (and it is per journey). So If you don't want this PR. We can just not do it.

skinkie:
> We need to discuss this: is the FareProductPrice a really good place for a quota ? Also, what does the quota applies to (the full FARE PRODUCT, a specific JOURNEY to which the product gives access to, etc.). Up to now we have the FareQuotaFactor and you can assign it via a ValidityPartameterAssignment to whatever it concerns.

I wonder if a quota is also something like should appear in SIRI similar to occupancy? Or is that really an out-of-the-box thought?

ue71603:
@skinkie  It came from the OJPFare2Nova example. The quotas are there. If SIRI has a Fare part we certainly can do it there. But I am not sure, where SIRI is considered with Fares. So I went to NeTEx. But  Christophe is not in favor (at least for the simple solution).

ue71603:
Nick: Can be done already with FareQuota --> Price and not this way around. 
(a) FareQuotaFactor
(b) number of units
(c) have a price.

trurlurl:
Removing "needs documentation update" and "needs transmodel update" since closed without merging.

Issue #593, opened 2023-12-13
Remove all External* elements in next


skinkie:
As discussed in #552, privateCodes and keylists are _available_.

ue71603:
asPR is open close here

PR #594, opened 2023-12-13
Remove External entities with for VDV compatibility in favor of Priva…

skinkie:
@duexw only if you like it.


skinkie:
@ue71603 @trurlurl need example updates.

skinkie:
Fix #593

ue71603:
I think we have to introduce PrivateCode in one or two places.
or have should I model:
`								<LineInDirectionRef>
									<ExternalDirectionRef type="uicType" ref="54"/>
									
									
								</LineInDirectionRef>
								<ScheduledStopPointRef version="any" ref="uic:007300101"/>
`

And how to do this one?

`						<Operator created="2003-06-09T14:20:00-05:00" changed="2004-05-09T14:20:00-05:00" modification="revise" version="1.0" id="noc:SCWW">
							
							<PublicCode>SCCW</PublicCode>
							<PrivateCode type="nocn">nocn:137725</PrivateCode>
							<ExternalOperatorRef type="vosa:LicenceNumber" ref="vosa:PD0000479"/>
							
							<Name>Stagecoach</Name>
							<ShortNa
`

skinkie:
@ue71603 I'll update all examples with #597 first.

Aurige:
didn't we say deprecate, not remove ? removing will be an issue for @duexw and probably others ...

skinkie:
@Aurige the point is most of these changes to next where never in master.

skinkie:
For filters we need better examples.

ue71603:
@skinkie  we will have to do this together. Not clear to me, how we should fix the examples.

ue71603:
@Aurige @nick-knowles @skinkie  @duexw  does this still work for you?

Aurige:
Prefer DEPRECATION (for the ones in master only) and recommend the use of external codes
Remove the ones only added in next...

Proposition to be updated

trurlurl (approved):

nick-knowles:
THis affects part1 docmentation for some core framework elemente.  UML has beenc ahnged . Note also  that PrivateCode should be moved from from Gereic utility types package to  to Responsibility model types package 
UML 
image

image


image

image

duexw:
I agreed on deprecating the ExternalXXXXRefs and using privateCodes instead. Alas I did not check if the privateCodes Element is always present where the ExternalXXXXRef has been dropped. For instance here:
image

If I used ExternalLineRef and ExternalDirectionRef before, I cannot replace it as there is no privateCodes Element. Do I need to raise an issue for this?

PR #595, opened 2023-12-13
Update description multi lingual string next

skinkie:


ue71603 (approved):

Aurige (approved):

nick-knowles:
Hi STefan

I thought we very specifically agreed that Codespace description  should  not be a Multilingual  string - I said that three times over!  Sorry if it was not clear. 
Making it  a complex type rather than a simple string  introduces an unnecessary dependency that I especially think we should avoid  . 
It is a metadata entry in the core model  that is not visible to end users and does not need translating.Can you make it a  string again please

The same principle  goes for the documentation on  delta trace but I am less fussed about that as

BTW Making it multistring also means The file no longer validates -  if there please try and keep files  independently validable.
e
Why  is it a bad idea? because  if we revise Multilingual string to allow alternative texts etc we force all that further uneccesary complexity (and possibly  further dependencies on  other modules) onto a kernel element. CODE SPace is an exception in other  ways - eg  it i isnt versioned

skinkie:
@nick-knowles can you please give us an example of a file that now does not validate? Or do you mean something else?

Issue #434, opened 2023-03-22
Prerequisites to usage of a line: you need to be in a car (e.g.)


ue71603:
Matthias:
We currently study NeTEx / SIRI/ OJP for car transport (on trains and ships)
 
Questions:
1.	How do we model this on a Line/VehicleJourney? With Mode/Submode it is not necessarily clear, that there are preconditions to transport. E.g. one needs a bicycle, bike, car or truck to get transported.
2.	Same applies to SIRI.
3.	For car transport one usually needs to be with a car (sometimes bike and sometimes even bicycle). How should we model this in the OJP request? Wit ItModeToCover and then it finds it? We also would need an ItModeToCover in OJPStopEventRequest and we need that information also in the response. Was something like that already considered?
 

Christophe:
There is a carTransportRailService submode ... and also a carTransporter TrainElementType. But I don't think that we really have an explicit "precondition".
Trucks are for freight, so not really in the scope of NeTEx
Bike in train is a facility or a BikeRack Equipement.

Matthias:

I don’t think this is what this is about.

We have a carTransportRailService between Kandersteg and Goppenstein. You can use it:
•	When you are driving a car
•	When you are driving a bike
•	When you are driving a bicycle
•	Trucks can only be loaded on trains with a special attribute (high-profile)
•	If you show up without a mentioned vehicle, then you can’t use it.

The green marked ones are sometimes not available for other carTransportRailServices in Switzerland.

For touristic transport sometimes you need skis, snowboard or a bicycle to use it.

So I guess we should define something in Transmodel and NeTEx that 

<PrerequisiteEquipmentOrVehicleList>car bike bicycle</PrerequisiteEquipmentOrVehicleList>
Or
<PrerequisiteEquipmentOrVehicleList>ski snowboard</PrerequisiteEquipmentOrVehicleList>
In the LinePropertiesGroup.

ue71603:
I think this is fixed with #600 as well. Not as a list, but almost similar. I guess if  @Aurige, if you prefer a List pls tell in #600

Aurige:
Doc updated based on https://github.com/NeTEx-CEN/NeTEx/pull/600

PR #600, opened 2023-12-13
Make sure that the PassengerCategoryies and the PrivateMode are done in NeTEx, so that it works out.

ue71603:
fixes #349


ue71603:
addresses #434 as well

xsd/netex_part_1/part1_networkDescription/netex_line_version.xsd
```diff
@@ -498,6 +498,31 @@ Rail transport, Roads and Road transport
...
+				<xsd:annotation>
+					<xsd:documentation>USER PROFILEs that are allowed on this line. Omitted means all.</xsd:documentation>
+				</xsd:annotation>
+				<xsd:complexType>
+					<xsd:sequence>
+						<xsd:element ref="UserProfileRef" minOccurs="0" maxOccurs="unbounded"/>
+					</xsd:sequence>
+				</xsd:complexType>
+			</xsd:element>
+			<xsd:element name="accessModes" minOccurs="0">
```
skinkie, line 511: Please do the enumeration list you mentioned in the call.

ue71603:
https://wiki.openstreetmap.org/wiki/Tag:bicycle%3Ddismount

ue71603:
RequiredAccessModes

ue71603:
different group so that is part of LineGroup

ue71603:
create a group and add it also to ServiceJourney.

ue71603:
Nick: Site has an element like that.

skinkie (approved):

Aurige (approved):

nick-knowles:
There is a small dependency hygiene issue here that has been overlooked.
**_UserType_** enums are  in Part3   (being created as a concept).  **_Line_** in Part2, so this gives a forward  dependency  from Part2 on Part3, something we normally strictly avoid.

A  Straightforward  way to resolved would be to move the UserTypeEnum to the ServiceRestrictions package in Part1 which contains various values relevant for describing both fares and journeys

PR #601, opened 2023-12-13
Integrate LineEquipment air conditioning

### LINKED ISSUE #480: Integrate LineEquipment air conditioning
We would like to integrate the line equipment data on the netex. 
It seems that the Netex profile for networks is appropriate, however only the notion of accessibility equipment is available. If we wish to integrate a data for example the presence of air conditioning on board a line (True or False) how can we integrate it into this netex profile?

ue71603:
fixes #480

### DIFF
--- a/xsd/netex_framework/netex_reusableComponents/netex_facility_support.xsd
+++ b/xsd/netex_framework/netex_reusableComponents/netex_facility_support.xsd
@@ -358,6 +358,21 @@ Rail transport, Roads and Road transport
 			<xsd:enumeration value="other"/>
 		</xsd:restriction>
 	</xsd:simpleType>
+	<xsd:simpleType name="EnvironmentControlEnumeration">
+		<xsd:annotation>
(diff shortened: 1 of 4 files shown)


xsd/netex_framework/netex_reusableComponents/netex_facility_support.xsd
```diff
@@ -358,6 +358,21 @@ Rail transport, Roads and Road transport
 			<xsd:enumeration value="other"/>
 		</xsd:restriction>
 	</xsd:simpleType>
+	<xsd:simpleType name="EnvironmentControlEnumeration">
```
ue71603, line 361: Nick: ClimateControl

Aurige (approved):

skinkie (approved):

nick-knowles:
UML Documentation added. 

NOTE : Should also add CLimateControl and LightingControl to NeTEx Accommodation element as facility may be specific to an accomodation type, eg 1st class berth might have blinds and aircon.

Netex Facility  values UML
image

Netex Accommodation model UML
image

PR #617, opened 2023-12-19
TEST: Remove seq seq

ue71603:


ue71603:
[Dok1.pdf](https://github.com/NeTEx-CEN/NeTEx/files/13714734/Dok1.pdf)

The removal of the inner sequence makes a difference (see attached word). However, It is unclear to me, why   AlternativeName_VersionedChild and AlternativeName contain the same content and a restriction was used for AlternativeName. Probably I miss the core principle for such inheritance.


I guess the reason is the extension – extension – restriction

In example 14.3 here: [Definitive XML Schema (1st edition) Examples: Deriving Complex Types (xs:restriction, xs:extension) (datypic.com)](http://www.datypic.com/books/defxmlschema1/chapter14.html)

It notes the effective model. So I guess the extension really added the sequence and therefore it must be there.

ue71603:
will not be done

PR #623, opened 2023-12-22
fixing imports to make files validate (multilingualstring)

ue71603:
Fixing #595 validation

@nick-knowles This doesn't solve the base discussion.


Aurige:
The Version will be deprecated, so only the Delta remains with the Multilangual Description. **IF** it does validate (XSD and XML examples) without issue, we can let it as it is (if not come back to xsd:string)

trurlurl (approved):

skinkie (approved):

ue71603:
@Aurige Are changes needed here?

PR #615, opened 2023-12-16
Deckplan seating plan (redo of 571, 574 and 575)

ue71603:
redo of #571  #574  and #575 

Pay attention there is something that is not working. I had to comment two lines out from #574. 
image
in xsd/netex_part_3/part3_fares/netex_accessRightParameter_version.xsd

A deep check by @Aurige  and @nick-knowles  is necessary

I did not recheck  #415  , #414, #413 and #412 but just closed them.

575 contained 
fa43f0f
0eddb08
29ee965
532c9b

574 contained 
"4247f2b
3b2cac2"
0eddb08
29ee965
532c9b

571 contained
0eddb08
29ee965
532c9be

PR #625, opened 2023-12-22
Enhancement -Site levels and GroupOfSites

nick-knowles:
Add SITE LEVEL STRUCTURW and LEVEL IN STRUCTURE as per Transmodel so as to support alternative label sets for levels on different building parts within a SITE
Update comment on RelativeLevelOrder attribute  previously added by Matthias to reflect enhancement  -ie that RelativeLevelOrder is now a 'site default' but that different alternative labbing sets are possible.

UML diagram
image


ue71603:
Last commit added to #615. Therefore I close this one.

Issue #556, opened 2023-11-10
Site Entrance qualifier 


Aurige:
Shouldn't we have an qualifier on SiteEntrance to say that it is a main Entrance/Exit, secondary
 Entrance/Exit, service Entrance/Exit or Emergency Entrance/Exit ?
Probably to be added to SiteEntranceGroup
![image](https://github.com/NeTEx-CEN/NeTEx/assets/3328999/bb3010e0-cd1b-4453-bbf7-f3e0f63a4aee)

skinkie:
I don't know how it is in France, but in The Netherlands I would consider even the back entrances of a stations to be 'very prominent'. Obviously there are less important entrances, but if you take a journey planner, wouldn't you want to enter the station were it is _close by_? Can you give a use case what this solves?

Aurige:
For journey planning, when you don't know yet from where you will arrive, it's often that the main entrance is used as destination for a site (unless the journey planner supports multiple destination points, of course). Also when drawing a map, you usually have a bigger pitcogram for the main entrance. But I agree with you the closest is always the best.

nick-knowles:
Relative importance /purpose of entrance could be added as a further EntrancePurpose enum .  Note that various other classifications  are already defined on SiteElementProperties, such as whether it is for public or private or staff us e only

Proposed revision

![image](https://github.com/NeTEx-CEN/NeTEx/assets/1826507/d138d996-36e7-4567-a393-17ca7c5032d5)

Aurige:
just remove Staf value

ue71603:
Through shop is secondary. Entrance has AvailabilityCondition. only enumeration. and attribute. 
@trurlurl will do a first draft

ue71603:
@trurlurl  is there anything in this issue that is not in the PR? If not, I suggest to close the issue (not the PR).

trurlurl:
> @trurlurl is there anything in this issue that is not in the PR? If not, I suggest to close the issue (not the PR).

Solved by #633, but not merged yet.

trurlurl:
I see that we don't want to wait for the merge of the PR, so I close the issue.

Aurige:
Part 1 updated (following https://github.com/NeTEx-CEN/NeTEx/pull/633)

PR #632, opened 2024-01-02
Site Entrance qualifier

trurlurl:
Intended to resolve issue #556.

Note that entrances can only be qualified as emergency exits if they are normally closed. If emergency routes needed to be described, a different mechanism would be required.

### DIFF
--- a/xsd/netex_part_1/part1_ifopt/netex_ifopt_site_support.xsd
+++ b/xsd/netex_part_1/part1_ifopt/netex_ifopt_site_support.xsd
@@ -336,6 +336,28 @@ Rail transport, Roads and Road transport
 			<xsd:enumeration value="other"/>
 		</xsd:restriction>
 	</xsd:simpleType>
+	<xsd:simpleType name="EntrancePurposeEnumeration">
+		<xsd:annotation>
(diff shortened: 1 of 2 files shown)


ue71603 (approved):
Should we add an example?

Aurige (approved):

skinkie:
> Should we add an example?

Yes please, helps validation.

skinkie:
This pull request targets master, is that intended?

trurlurl:
> This pull request targets master, is that intended?

Yes - would there have been "better intentions"?

skinkie:
> > This pull request targets master, is that intended?
> 
> Yes - would there have been "better intentions"?

It is not a fix, it is not a documentated change... I would argue _next_, since EPIAP is also finished.

trurlurl:
> > > This pull request targets master, is that intended?
> > 
> > 
> > Yes - would there have been "better intentions"?
> 
> It is not a fix, it is not a documentated change... I would argue _next_, since EPIAP is also finished.

I've done it in a new PR - #633.

ue71603:
Then I close this one...

PR #633, opened 2024-01-02
Site Entrance qualifier (on next)

trurlurl:
Same as #632, but for next instead of master. To resolve issue #556.

Additional change in the annotation text for EntranceEnumeration.


skinkie:
@trurlurl can we modify a single example so a value is used?

trurlurl:
> @trurlurl can we modify a single example so a value is used?

@skinkie Of course, sorry, I'll do it tomorrow,.

skinkie (approved):
Good example!

ue71603 (approved):

Aurige (approved):

PR #638, opened 2024-01-17
Fix passenger category for line modularisation

nick-knowles:
Modularisation was wrong. Part1 dependency on Part3. Fix by moving Usertypes from netex_line_support.xsd to part1 netex_serviceRestrictions file. 
Also 
(a) add "veteran" as user type 
(b) reorder new attributes so that restrictedLined is  next to UserTypes 
(c) improve comments (d) add example to test


nick-knowles:
Created on wrong base

Issue #641, opened 2024-01-17
Deprecate the Version element


skinkie:
https://github.com/BISONNL/NeTEx-NL/issues/71

ue71603:
@skinkie Does this need a PR?

skinkie:
This should still be a pull request!

PR #642, opened 2024-01-17
Remove versions and Version-elements

skinkie:
Big change. Please review!


skinkie:
fix #641

ue71603 (approved):

duexw (approved):
I approve of the change in general. While we do use the version attribute on some elements, our implementation of the standard does not use the version element and we have not yet received data from others where it was used.  
We should think about withdrawing the "Netex versioning" xml example, as it may not make too much sense without the version element.

trurlurl (approved):

xsd/netex_framework/netex_responsibility/netex_versionFrame_version.xsd
```diff
@@ -117,11 +117,6 @@ Rail transport, Roads and Road transport
 					<xsd:documentation>Default values to use on elements in the frame  that do not explicitly state a value.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
-			<xsd:element name="versions" type="versions_RelStructure" minOccurs="0">
```
Aurige, line 120: didn't we agree to deprecate ? 
I  don't have my notes with me, but this may break quite a lot a existing data

skinkie: In _next_ we don't have to deprecate right? That is what should happen in master.

Aurige: that's not my understanding... we decided to use deprecate multiple times to limits the incompatibilities. And for version, the decision of the group that I wrote on my notes is "deprecate"

Aurige: no, my understanding is that the group wanted to deprecate on 'next' (we already did multiple deprecation on next) in order to give time to update production. We will then the remove in 3.0

Aurige: https://github.com/NeTEx-CEN/NeTEx/issues/641

skinkie: It is fine by me. We have assured in this pull request that all examples are stripped of the Version element, we can 'revert' the removed elements.

Aurige:
Keep examples updated, and put Version back as deprecated

PR #644, opened 2024-01-18
fix CI

juliustens-db:


juliustens-db:
Sorry for the spam, I accidentally created a PR to the main repo instead of our fork because Github always suggests that by default 😅

skinkie:
@juliustens-db we obviously allow improvements ;-) so please do ;-)

PR #657, opened 2024-02-07
change all order to positiveInteger,

ue71603:
add info about what to do what to do if order makes no sense (order="1" for all elements)

replaces https://github.com/NeTEx-CEN/NeTEx/pull/477/


skinkie:
@ue71603 out of the box thinking shall we introduce a simpleType elementOrder being positiveInteger, that can be applied?

ue71603:
@Aurige should I do it?

trurlurl (approved):

PR #555, opened 2023-11-10
EPIAP Equipments Enhancement

Aurige:
Enhancements of the Equipments proposed by the French team working on an accessibility capture tool

These enhancements are mainly coming from comparison and mapping with OSM
The original request is here etalab/transport-normes[#18](https://github.com/NeTEx-CEN/NeTEx/issues/18) (French Accessibility Profile)

The first commit ins on Crossing Equipments
More similar commits (on other accessibility related equipments) will follow on this same PR (several proposals on several objects on the same XSD file)

PR #664, opened 2024-02-13
`TactileWriting` attribute for all handrails

juliustens-db:
Hey everyone! #555 introduced a `TactileWriting` attribute to `StairGroup`, indicating if the handrail includes a tactile sign. IMO it makes sense to also include that attribute for other equipments which have a handrail, specifically `Ramp` and `Lift`.

Maybe it would make sense to move all handrail-specific attributes to a separate group, to make sure that they are always in sync. But that's probably out of scope for this small MR 😄.

### DIFF
--- a/xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_version.xsd
+++ b/xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_version.xsd
@@ -833,6 +833,11 @@ Rail transport, Roads and Road transport
 					<xsd:documentation>Height of lower handrail from floor.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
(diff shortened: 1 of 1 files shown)


skinkie (approved):
@ue71603 @Aurige can we generatise the Handrail stuff in a separate group?

skinkie:
@juliustens-db if you want to do the grouping that is fine too :-)

juliustens-db:
@skinkie Basically one could reuse the `LiftHandrailGroup`, which exactly contains those handrail-specific attributes. I was not sure however, if it would be fine to just rename the group, or if one would need to keep the old one for backwards-compatibility (I don't know if you consider XML groups to be part of the standard "API" or an internal implementation detail). Depending on the response I can implement it either way.

ue71603 (approved):

Joostb61 (approved):

trurlurl (approved):

PR #672, opened 2024-03-01
TM enhancement  PASSENGER AT STOP TIME with example

nick-knowles:
a) Add enhancement to allow a PASSENGER AT STOP TIME  to be added, to allow a time to be at station that is earlier than the departure time to be specified.   
b) NB when adding constraints for the new element.  I noticed there is no uniqueness constraint on the existing  **_TimetabledPassingTigme_**.  I have added the  missing TimetabledPassingTime uniqueness constraint and corrected the examples to have unique values.  Note that this may highlight uniquness errors in existing data sets.
c) Add a few extra values to CHECK CONSTRAINT process enums eg assistedBoarding / alighting and to the 
d) Extend the train DECK PLAN  example to have two examples of PassengerAtStop times.  (i) A simple earlier time to the  for assisted boarding, and (ii) a motorail loading time that is earlier than the departure . For the latter have added a motorail loading carriage to thetrain in the  DECK PLAN  example,  along with  a motorail loading stop place (with car only access mode)  This also required to add a new enumeration value openWhileLoading yo the doorsetting enumeration to cover the case of an entrance that is open just while loading.

d_ Also add a separate version of the train deck plane example stripped down to have just the seating  spots: this is intended to indicate the minimum DEck Plan data set needed to support booking and reservation without bothering with layout etc.


ue71603:
covered in #674

PR #673, opened 2024-03-01
TM Enhancement: add supplier CONTRACT entity and reference it from JOURNEY ACCOUNTING

nick-knowles:
Add CONTRACT for use on JOURNEY ACCOUNTING etc, 
Add uniquness constraint

Split  off some elements from responsibility_support into responsibilitySet_support

Update JOURNEY ACCOUNTING examples to use COntract


ue71603:
covered in °674

PR #685, opened 2024-03-19
Make OrganisationRef abstract

### LINKED ISSUE #679: Make OrganisationRef abstract
Since Organisation is an abstract concept, it does not really make sense to have OrganisationRef exist on its own, which is already used as a subsitutionGroup for the abstract OrganisationRef. Force the user to use AuthorityRef, OperatorRef and likes.

skinkie:
If we want to target master with this, we must deprecate _OrganisationRef_ I wonder if we can take the 'dummy route' here. Where OrganisationRef is deprecated, and OrganisationRef_ would be the dummy type.

Fix #679


trurlurl (approved):

ue71603 (approved):

Aurige (approved):

Aurige:
Note added in Part 1 (mentioning that Organisation is abstract) but this level of XSD is not really detailed in the documentation

PR #686, opened 2024-03-19
Change organisation ref abstract

skinkie:
This does the trick in master **with** backwards compatibility and an example to validate for the backwards compatibility. So master it is.


ue71603:
@duexw  What do you think?

trurlurl (approved):

ue71603 (approved):

xsd/NeTEx_publication_timetable.xsd
```diff
@@ -1285,17 +1285,6 @@ Provides a general purose wrapper for NeTEx data content.</xsd:documentation>
 			<xsd:field xpath="@id"/>
 			<xsd:field xpath="@version"/>
 		</xsd:key>
-		<!-- =====Organisation============================== -->
```
skinkie, line 1288: Investigate if this deletion actually is correct. I remember undoing it.

ue71603: it seems to me that it is is still in the file on line 1376ff

ue71603:
@Aurige  pls review

Aurige:
looks ok :-)

Aurige (approved):

PR #681, opened 2024-03-14
Feature deckplan

skinkie:
Everything from #574, #575, #625 
#674 spot affinity done. The rest probably is more bogus and is not done.

Will continue with #656

PR #695, opened 2024-03-26
D eck plan revisions

nick-knowles:
Further correction - allow simple PathLinks in SieFrame


ue71603:
will all be done in #681

PR #696, opened 2024-03-28
Add properties to describe the height of a Lift

juliustens-db:
Hey everyone! At Deutsche Bahn, we (sometimes) have records on the height of our elevator cabins and doors, and figured it would be nice to include that data in our new NeTEx datasets (which we are currently working on).

I therefore created this PR, which simply clones the `Width` attributes used to describe a lift equipment's cabin and door width. If you'd prefer another approach or different names, I'll update the PR as needed 🙂

### DIFF
--- a/xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_version.xsd
+++ b/xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_version.xsd
@@ -156,6 +156,11 @@ Rail transport, Roads and Road transport
 					<xsd:documentation>Width of EQUIPMENT or entrance to EQUIPMENT (e.g. Lift).</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
+			<xsd:element name="Height" type="LengthType" minOccurs="0">
+				<xsd:annotation>
+					<xsd:documentation>Height of EQUIPMENT or entrance to EQUIPMENT (e.g. Lift).</xsd:documentation>
+				</xsd:annotation>
+			</xsd:element>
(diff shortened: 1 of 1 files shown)


skinkie:
Please target master and/or next. EPIAP is already merged in both.

juliustens-db:
@skinkie thanks, I updated the PR. Dumb question: Why does the EPIAP branch still exist? 😅

skinkie:
> @skinkie thanks, I updated the PR. Dumb question: Why does the EPIAP branch still exist? 😅

I guess the true question is, should it remain, since it is 'sort of' a release.

juliustens-db:
> I guess the true question is, should it remain, since it is 'sort of' a release.

IMO it would be better to use a tag for that, so that noone can accidentally push commits to it

skinkie:
> Also, is it really merged already? I can't find the commits on master (or I am blind 😄)

@ue71603 what was the situation? We did it only on next because it was still in voting? #help

> > I guess the true question is, should it remain, since it is 'sort of' a release.
> 
> IMO it would be better to use a tag for that, so that noone can accidentally push commits to it

There is a different idea here. If we would release a major version the idea would be that fixes could still go in to that branch. For feature releases, the feature branch should be indeed removed after it was merged.

ue71603:
@skinkie  why can't I write a review? Also. Is EPIAP not only merged into next yet.   master=1.2.3, EPIAP = release 1.3, next= release 2.0

skinkie:
If EPIAP voting is done, we can go for master.

ue71603 (approved):

trurlurl (approved):

ue71603:
@skinkie  EPIAP was accepted. Merge?

ue71603:
do master first.

skinkie (approved):

Issue #665, opened 2024-02-14
`TactileActuators` attribute on Escalators and Travelators


juliustens-db:
Hey everyone! Escalator and Travelator objects have an attribute called `TactileActuators`, with the description "wether [element] has a tactile actuator". People who build these equipments probably know exactly what that means, but as someone who is "just" implementing the standard, I struggle to understand the usage of this property. Could you provide some details on the usage of such an actuator in practice? (What does it control/activate?). Sorry if this is obvious and I'm just missing the point 😄

ue71603:
@trurlurl  @Joostb61 Can we update the definition?

Joostb61:
This is the description in the EPIAP functional document:
Whether there are buttons to start the escalator if it is not moving, and whether these buttons are intended and suited for the visually impaired.

ue71603:
fixed in #720

juliustens-db:
Thanks, that helps a lot! 🙂

PR #705, opened 2024-04-15
EPIAP to Master

skinkie:

PR #720, opened 2024-05-02
Update neTactileActuators attribute on Escalators and Travelators #665tex_ifopt_equipmentAccess_version.xsd

ue71603:
fixes: https://github.com/NeTEx-CEN/NeTEx/issues/665

### DIFF
--- a/xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_version.xsd
+++ b/xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_version.xsd
@@ -956,7 +956,7 @@ Rail transport, Roads and Road transport
 			</xsd:element>
 			<xsd:element name="TactileActuators" type="xsd:boolean" minOccurs="0">
 				<xsd:annotation>
-					<xsd:documentation>Whether there are buttons to start the TRAVELATOR if it is not moving, and whether these buttons are intended and suited for the visually impaired.</xsd:documentation>
+					<xsd:documentation>Whether all buttons have tactile markers suited for the visually impaired.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
 			<xsd:element name="AudioAnnouncements" type="xsd:boolean" minOccurs="0">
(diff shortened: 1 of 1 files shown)


xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_version.xsd
```diff
@@ -956,7 +956,7 @@ Rail transport, Roads and Road transport
 			</xsd:element>
 			<xsd:element name="TactileActuators" type="xsd:boolean" minOccurs="0">
 				<xsd:annotation>
-					<xsd:documentation>Whether Lift has a tactile actuator.</xsd:documentation>
+					<xsd:documentation>Whether there are buttons to start the TRAVELATOR if it is not moving, and whether these buttons are intended and suited for the visually impaired.</xsd:documentation>
```
trurlurl, line 959: No, these here are the *TactileActuators* of the **LiftEquipment**. The change for Escalator and Travelator is already done (cf. line 726), I introduced it as a commit in #705.

trurlurl (changes requested):

xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentAccess_version.xsd
ue71603, line 959:
```suggestion
					<xsd:documentation>Whether there are buttons to start the LIFT if it is not moving, and whether these buttons are intended and suited for the visually impaired.</xsd:documentation>
```

ue71603: better with the new commit? @trurlurl

trurlurl: Given that in the *LiftButtonsGroup* we have *BrailleButtons*, *RaisedButtons*, and *TactileGroundFloorButton*, then the only meaning that seems to make sense for me would be something like: "Whether at least the call buttons have tactile markers for the visually impaired." OR **"Whether all buttons have tactile markers suited for the visually impaired."** The second one sounds better to me, but I don't know what the initial intention was.

trurlurl (approved):

skinkie (approved):

PR #746, opened 2024-06-27
Checking for all examples

ue71603:
In reality probable some examples correctly should no longer work. VDV 452 examples are not in a NeTEX 2.0 format.

### DIFF
--- a/.github/scripts/validate-examples.sh
+++ b/.github/scripts/validate-examples.sh
@@ -9,7 +9,7 @@ set -e
 
 echo "Validating NeTEx XML examples ..."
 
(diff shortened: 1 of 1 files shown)


.github/scripts/validate-examples.sh
```diff
@@ -9,7 +9,7 @@ set -e
...
+if xmllint --noout --schema xsd/NeTEx_publication.xsd examples/standards/*xml examples/standards/*/*xml examples/standards/*/*/*xml && xmllint --noout --schema xsd/NeTEx_publication.xsd examples/functions/*xml examples/functions/*/*xml examples/functions/*/*/*xml examples/functions/*/*/*/*xml; then
```
skinkie, line 12: This fails due to xmllint breaking on epiap.

skinkie:
We want this one if the xmllint stuff is resolved;
`find examples -name "*.xml" -type f -print0 | xargs -0 xmllint --noout --schema xsd/NeTEx_publication.xsd`

Aurige:
I'm a bit puzzled, because some files that are failing here (like tap_tsi\B3_Special\Netex_tap_tsi_B3+more.xml ) are validating on next for me

skinkie:
@Aurige I think in principle we must fix the examples in _master_ first.

Aurige:
Yes that's of course fine to fix them in master, however I don't understand why the first error (and most of the others) is 

 _examples/standards/efp/VVS_Fares_V1-ZoneSeries-nk3.xml:585: element StakeholderRoleType: Schemas validity error : Element '{http://www.netex.org.uk/netex}StakeholderRoleType': [facet 'enumeration'] The value '**_EntityLegalOwnership_**' is not an element of the set {'planning', 'operation', 'control', 'reservation', '**_entityLegalOwnership_**', 'fareManagement', 'financing', 'securityManagement', 'customerService', 'dataRegistrar', 'tenant', 'facilityManagement', 'lessor', 'dataRegistrar', 'other'}._

But even if deprecated, **_EntityLegalOwnership_** does now exist and should not raise an error (the list of possible values in the error message is not the right one currently in next)

skinkie:
Check if we can install https://launchpad.net/ubuntu/plucky/+source/libxml2 in CI.

ue71603:
cool

Von: Stefan de Konink ***@***.***>
Gesendet: Freitag, 22. November 2024 11:22
An: NeTEx-CEN/NeTEx ***@***.***>
Cc: Günter Matthias (IT-PTR-EXT-EXT2 - Extern) ***@***.***>; Author ***@***.***>
Betreff: Re: [NeTEx-CEN/NeTEx] Checking for all examples (PR #746)


Check if we can install https://launchpad.net/ubuntu/plucky/+source/libxml2 in CI.

-

skinkie:
Will be fixed by #916

PR #743, opened 2024-06-26
Prioritize TAP TSI validation

skinkie:

PR #754, opened 2024-07-16
TAP TSI Example validation

Aurige:
There were some real errors in some examples that were not catched before (reused ID, bad references...), and also a missed deprecation on enumeration values. Correcting these seems to solve the TAP TSI example validation issues.

Should solve https://github.com/NeTEx-CEN/NeTEx/pull/743


skinkie:
This should go into master?

Aurige:
not for the deprecations ... but we can also update the examples in master

skinkie (approved):

ue71603 (approved):

Issue #757, opened 2024-07-25
OccupancyInFrame


skinkie:
As an addition of #605 so it can be independent of the Timetable, and the Timetable can reference to it.

ue71603:
We have OccupancyViewInFrameGroup

PR #605, opened 2023-12-13
Predicted occupation

ue71603:
solves #398 

However, currently only works for Calls

@Aurige I need guidance: How to put this into PassingTime /VDV 462 method. And also do we need for new mode a version that puts it on the Vehicle? Somewhere else?

Also folders: Should the thing be in a different folder? Should other NeTEx specific elements be incorporated?
Did I get the right "entrance" into formation?

PR #784, opened 2024-09-30
OccupancyView also possible directly in ServiceFrame

ue71603:
fixes https://github.com/NeTEx-CEN/NeTEx/issues/757


examples/functions/timetable/Netex_01.1_Bus_SimpleTimetable_JourneysOnly.xml
```diff
@@ -220,6 +220,17 @@ The Calendar is shown coded as
 							<VehicleModes>bus</VehicleModes>
 						</ScheduledStopPoint>
 					</scheduledStopPoints>
+					<occupancies>
+						<OccupancyView version="any" id="occinFrame">
```
skinkie, line 224: I don't see how a Occupancy**View** at this point. I think it must not be a _view_.

ue71603: No other elements available. It always was a View and never anything else. You wanted it here. Sorry.

skinkie: No I want to have an _Occupancy_ here.

ue71603:
There exists currently only OccupancyView. So what do you want? Renaming? Wrapper?

ue71603:
Then this is no longer a small PR, but a full fledged one. And the question can be asked, why it supports only a part of what is possible in SIRI.

Aurige:
The group discussion was for an  OccupancyView (cf https://github.com/NeTEx-CEN/NeTEx/pull/605 ) so that looks fine to me ... Shall we discuss it again ?

skinkie:
I was in that meeting, and the idea was to make Occupancy a first class object, not a view.

Aurige:
The Full TM compliant Occupancy was identified as too rich and not fulfilling our more simple requirements, so a SIRI inspired view was decided ... but you can draft a fully TM compliant Occupancy for next release

ue71603:
@skinkie  currently set for 3.0. Do you need it before?

Aurige:
is there any possible ref from Jouneys to that OccupancyView ?
More discussion needed

ue71603:
can be addressed in 3.0

Issue #748, opened 2024-07-02
HeadwayJourneyGroupGroup introduces a Description which is already present in GroupOfEntitiesGroup


skinkie:
https://github.com/NeTEx-CEN/NeTEx/blob/master/xsd/netex_part_2/part2_journeyTimes/netex_vehicleJourneyFrequency_version.xsd#L294

https://github.com/NeTEx-CEN/NeTEx/blob/master/xsd/netex_framework/netex_genericFramework/netex_grouping_version.xsd#L108

warning: Converting parent field `{http://www.netex.org.uk/netex}GroupOfEntities_VersionStructure::Description` to a list to match child class `{http://www.netex.org.uk/netex}HeadwayJourneyGroup`

Removing the extra Description in HeadwayJourneyGroupGroup resolves the issue.

In Next the same would go for PathLinkInSequenceGroup, removing the extra Description resolves the issue.

ue71603:
@skinkie ok for me to remove it. Two PR? one to master and one to next?

skinkie:
I have to check if the PathLinkInSequenceGroup is also in master.

skinkie:
This is the same thing as #562

ue71603:
@skinkie  PR to remove it?

Aurige:
If that does not break examples (which is most most likely) I would remove this extra Description

ue71603:
PR done

skinkie:
@ue71603 will you do it for PathLinkInSequenceGroup as well?

Aurige:
Doc updated via https://github.com/NeTEx-CEN/NeTEx/pull/783

PR #785, opened 2024-09-30
remove description in PathLinkInSequenceGroup

ue71603:
would be duplicated otherwise

### DIFF
--- a/xsd/netex_part_1/part1_ifopt/netex_ifopt_navigationPath_version.xsd
+++ b/xsd/netex_part_1/part1_ifopt/netex_ifopt_navigationPath_version.xsd
@@ -362,11 +362,6 @@ Rail transport, Roads and Road transport
 		</xsd:annotation>
 		<xsd:sequence>
 			<xsd:element ref="PathLinkRef"/>
-			<xsd:element name="Description" type="MultilingualString" minOccurs="0">
-				<xsd:annotation>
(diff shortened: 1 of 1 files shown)


ue71603:
@skinkie for this not possible as description was used. close?

skinkie:
> @skinkie for this not possible as description was used. close?

What you mean in this case with 'used'? Are there two descriptions used at the same place?

ue71603:
Hi Stefan
Removing it in master resultated in many invalid examples (see ci /run). So there is no other one (or at least it is not at exactly the same place).

ue71603:
The problem is, that the Description now would need to be before the PathLinkRef

it was

* Description
* PathLinkRef
* Description

So the position of the Description would need to be changed in all examples.

image

skinkie:
The suggestion was to remove it from _next_.

ue71603:
no, was not: you referenced master in https://github.com/NeTEx-CEN/NeTEx/issues/748

skinkie:
I am happy to update all the examples. But I wonder who uncarefully duplicated, provided the wrong examples (and caused this mess).

skinkie:
> no, was not: you referenced master in #748

_In Next the same would go for PathLinkInSequenceGroup, removing the extra Description resolves the issue._

ue71603:
I move the PR to next and update the examples.

Issue #772, opened 2024-09-04
Confused by VehicleMode, TransportMode, AllVehicleModesOfTransport, etc.


trurlurl:
There are elements like _VehicleMode(s), TransportMode(s), OtherTransportModes, Mode_ that map to the following three enumerations, but in very inconsistent or non-obvious ways: _**VehicleModeEnumeration, AllVehicleModesOfTransportEnumeration, AllModesEnumeration**_. It would be helpful to disentangle things a bit - by improving annotations and perhaps, where possible, by renaming elements and the three enumeration types.

What I've found (unless I made some mistake):
- _AllVehicleModesOfTransportEnumeration_ is a superset of _VehicleModeEnumeration_, with additional values _unknown, anyMode, intercityRail, urbanRail_
- _AllModesEnumeration_ is a superset of _AllVehicleModesOfTransportEnumeration_, with additional values _foot, bicycle, motorcycle, scooter, car, shuttle_

I guess we can't get rid of _VehicleModeEnumeration_ in favor of _AllVehicleModesOfTransportEnumeration_ - or can we? If not, it would be good to describe the difference / rationale.

Could we choose **less confusing names?** E.g., _PrimaryVehicleModesOfTransportEnumeration, AllVehicleModesOfTransportEnumeration, AllModesOfTransportEnumeration_?

If the subset-superset hierarchy between the three enumerations were to hold in the future - would it make sense to inherit the values from the subset using the _union_ construct in XSD?

As for the **mapping from element names to the enumerations**, could we choose more consistent names? 

- I guess that _VehicleMode_ needs to stay for backwards compatibility. But then it would be good if _OtherTransportModes_ (which also maps to _VehicleModeEnumeration / PrimaryVehicleModesOfTransportEnumeration_) could be renamed to _OtherVehicleModes_ or _OtherPrimaryVehicleModes._

- _TransportMode(s)_ should always map to _AllModesEnumeration / AllModesOfTransportEnumeration_.

- A new name _VehicleTransportMode_ might be introduced to consistently map to _AllVehicleModesOfTransportEnumeration_?

Last question: the plural forms referring to a list should, in principle, start with a lower case (_transportModes_). Change or leave it (many places)?

Depending on your feedback I would prepare a PR.

ue71603:
@Aurige  Can you pls instruct André? Or we discuss it tomorrow.

Aurige:
[...]

PR #791, opened 2024-10-18
Stratification of transport mode enumerations

trurlurl:
This resolves issue #772. Stratified `VehicleModeEnumeration, AllVehicleModesOfTransportEnumeration, AllModesEnumeration` to ony two: `AllVehicleModesOfTransportEnumeration` and `AllModesOfTransportEnumeration`.

`VehicleMode` now uniformly maps to `AllVehicleModesOfTransportEnumeration`.
`TransportMode` maps to both `AllVehicleModesOfTransportEnumeration` and `AllModesOfTransportEnumeration`, but context and annotations prevent confusion, in my view. 

The changes should keep backwards compatibilty, though I changed the mappings (to an enumeration type) at a few places where I thought the chosen enumeration likely to be erroneous. I'm going to mark them in this PR.

### DIFF
--- a/xsd/netex_framework/netex_reusableComponents/netex_access_version.xsd
+++ b/xsd/netex_framework/netex_reusableComponents/netex_access_version.xsd
@@ -148,9 +148,9 @@ Rail transport, Roads and Road transport
 			<xsd:documentation>Type for ACCESS link end.</xsd:documentation>
 		</xsd:annotation>
 		<xsd:sequence>
-			<xsd:element name="TransportMode" type="AllModesEnumeration" default="all" minOccurs="0">
+			<xsd:element name="TransportMode" type="AllModesOfTransportEnumeration" default="all" minOccurs="0">
 				<xsd:annotation>
-					<xsd:documentation>Identifier of MODE of end point of ACCESS link. Default is all modes.</xsd:documentation>
+					<xsd:documentation>Identifier of MODE of TRANSPORT of end point of ACCESS link. Default is all modes.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
 			<xsd:element name="PlaceRef" type="PlaceRefStructure" minOccurs="0">
--- a/xsd/netex_framework/netex_reusableComponents/netex_mode_support.xsd
+++ b/xsd/netex_framework/netex_reusableComponents/netex_mode_support.xsd
@@ -112,41 +112,13 @@ Rail transport, Roads and Road transport
 		<xsd:list itemType="AccessModeEnumeration"/>
 	</xsd:simpleType>
 	<!--======Vehicle===================================================================================-->
-	<xsd:element name="VehicleMode" type="AllModesEnumeration">
+	<xsd:element name="VehicleMode" type="AllVehicleModesOfTransportEnumeration">
 		<xsd:annotation>
-			<xsd:documentation>VEHICLE MODE: a characterisation of the operation according to the means of transport (bus, tram, metro, train, ferry, ship).</xsd:documentation>
+			<xsd:documentation>VEHICLE TRANSPORT MODE: a characterisation of the operation according to the means of transport (bus, tram, metro, train, ferry, ship).</xsd:documentation>
 		</xsd:annotation>
 	</xsd:element>
-	<xsd:simpleType name="VehicleModeEnumeration">
-		<xsd:annotation>
-			<xsd:documentation>Allowed values for MODES of Public Transport applicable to timetabled public transport.</xsd:documentation>
-		</xsd:annotation>
-		<xsd:restriction base="xsd:NMTOKEN">
-			<xsd:enumeration value="air"/>
-			<xsd:enumeration value="bus"/>
-			<xsd:enumeration value="coach"/>
-			<xsd:enumeration value="ferry"/>
-			<xsd:enumeration value="metro"/>
-			<xsd:enumeration value="rail"/>
-			<xsd:enumeration value="trolleyBus"/>
-			<xsd:enumeration value="tram"/>
-			<xsd:enumeration value="water"/>
-			<xsd:enumeration value="cableway"/>
-			<xsd:enumeration value="funicular"/>
-			<xsd:enumeration value="lift"/>
-			<xsd:enumeration value="snowAndIce"/>
-			<xsd:enumeration value="other"/>
-		</xsd:restriction>
-	</xsd:simpleType>
-	<xsd:simpleType name="VehicleModeListOfEnumerations">
-		<xsd:annotation>
-			<xsd:documentation>Allowed values for List of PT MODES of transport applicable to timetabled public transport.</xsd:documentation>
-		</xsd:annotation>
-		<xsd:list itemType="VehicleModeEnumeration"/>
-	</xsd:simpleType>
 	<!--======Modes===================================================================================-->
-	<!--======Modes===================================================================================-->
-	<xsd:element name="RoadVehicleMode" type="AllModesEnumeration">
+	<xsd:element name="RoadVehicleMode" type="RoadVehicleModeEnumeration">
 		<xsd:annotation>
 			<xsd:documentation>Road Vehicle MODE: a characterisation of the operation according to the means of transport (bus, tram, coach).</xsd:documentation>
 		</xsd:annotation>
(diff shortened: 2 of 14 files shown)


xsd/netex_framework/netex_reusableComponents/netex_transportOrganisation_version.xsd
```diff
@@ -731,7 +731,7 @@ Rail transport, Roads and Road transport
 			<xsd:documentation>Elements for an TRANSPORT ADMINISTRATIVE  ZONE.</xsd:documentation>
 		</xsd:annotation>
 		<xsd:sequence>
-			<xsd:element name="VehicleModes" type="AllModesListOfEnumerations" default="all" minOccurs="0">
```
trurlurl, line 734: This looks like having been the wrong before.

xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentTicketing_version.xsd
```diff
@@ -114,9 +114,9 @@ Rail transport, Roads and Road transport
 			<xsd:documentation>Elements for a TICKETING EQUIPMENT.</xsd:documentation>
 		</xsd:annotation>
 		<xsd:sequence>
-			<xsd:element name="VehicleModes" type="AllModesListOfEnumerations" minOccurs="0">
```
trurlurl, line 117: This looks like having been the wrong before.

xsd/netex_framework/netex_reusableComponents/netex_mode_support.xsd
```diff
@@ -112,41 +112,13 @@ Rail transport, Roads and Road transport
 		<xsd:list itemType="AccessModeEnumeration"/>
 	</xsd:simpleType>
 	<!--======Vehicle===================================================================================-->
-	<xsd:element name="VehicleMode" type="AllModesEnumeration">
```
trurlurl, line 115: This looks like having been the wrong before.

xsd/netex_framework/netex_reusableComponents/netex_mode_support.xsd
```diff
@@ -112,41 +112,13 @@ Rail transport, Roads and Road transport
...
-	</xsd:simpleType>
-	<xsd:simpleType name="VehicleModeListOfEnumerations">
-		<xsd:annotation>
-			<xsd:documentation>Allowed values for List of PT MODES of transport applicable to timetabled public transport.</xsd:documentation>
-		</xsd:annotation>
-		<xsd:list itemType="VehicleModeEnumeration"/>
-	</xsd:simpleType>
-	<!--======Modes===================================================================================-->
 	<!--======Modes===================================================================================-->
-	<xsd:element name="RoadVehicleMode" type="AllModesEnumeration">
```
trurlurl, line 149: This looks like having been the wrong before.

skinkie (dismissed):
You get my approval because the tests / checks succeed not because I am at academic level aware if this change is "transmodel compliant" etc.

In all cases: simplification is a good job.

xsd/netex_framework/netex_reusableComponents/netex_access_version.xsd
Ulf9, line 1: If we are to make any changes we should try to align the suggested type-names and annotation as far as possible with what Transmodel 6.2 states to not further add to the confusion. I am not sure that the suggested changes do align with Transmodel 6.2.    We need to basically separate between the combination of conventional modes of operation with alternative modes of operation on one side i. e. "advertised transport" and personal modes of operation  e.g. with your own bike or your own car (when you are not car-sharing). Thus in Transmodel either PUBLIC TRANSPORT MODE or PERSONAL MODE.
If you wish to generalise  PUBLIC TRANSPORT MODE and PERSONAL MODE  you use the term MODE.

These three terms are defined in Transmodel as:

MODE :
A broad characterisation of the means of movement and / or operational aspect in consideration (e.g., by bus, by train, by metro, by ferry, by foot, by private car, etc.).

----------------
PUBLIC TRANSPORT MODE: 
A MODE used for public transport services (e.g., by bus, by tram, by metro, by train, by ferry, by ship, by shared taxi car, by shared car) under either a CONVENTIONAL MODE OF OPERATION or an ALTERNATIVE MODE OF OPERATION (i.e., excluding a PERSONAL MODE OF OPERATION).
Note: Prior to TM 6.2 this was named VEHICLE MODE
----------------
PERSONAL MODE :
A MODE used outside of public transport services (e.g., by private car, by private cycle, by private scooter, by private boat) under a PERSONAL MODE OF OPERATION.
----------------

Thus, please change the terminology of type-names and annotation accordingly and use capital letters only in accordance with Transmodel.

ue71603:
@trurlurl  Can you incorporate the input from Ulf?
Transmodel terms can be found here https://transmodel-cen.eu/model/

trurlurl:
Thank you @Ulf9 ! So, in order to match Transmodel we could 

(a) rename `AllVehicleModesOfTransportEnumeration` to `AllPublicTransportModesEnumeration` with the annotation: "PUBLIC TRANSPORT MODE: A MODE used for public transport services (e.g., by bus, by tram, by metro, by train, by ferry, by ship, by shared taxi car, by shared car) under either a CONVENTIONAL MODE OF OPERATION or an ALTERNATIVE MODE OF OPERATION (i.e., excluding a PERSONAL MODE OF OPERATION). Note: Prior to TM 6.2 this was named VEHICLE MODE.", and 

(b) rename `AllModesOfTransportEnumeration` to  `AllModesEnumeration` with annotation "All MODEs (PUBLIC TRANSPORT MODEs and PERSONAL MODEs): TPEG pti_table 01."

I would prefer, however, to be able to have, instead of (b): `AllTransportModesEnumeration`. But having read your explanations, I suspect that "TRANSPORT MODE" is not a Transmodel term and "TRANSPORT" only appears in "PUBLIC TRANSPORT" ? Such that "PUBLIC TRANSPORT MODE" is only a MODE of PUBLIC TRANSPORT and never a PUBLIC "TRANSPORT MODE". 

If you agree with that then I'm going to do the necessary changes.

Aurige:
> If you agree with that then I'm going to do the necessary changes.

That's fine for me ... looks pretty well aligned with Transmodel. 
Maybe skip the reference to tpeg since their codes are not stable (or just mention something like "initilised taking into account TPEG pti_table 01, with its 2010 version" ... but that may not really be helpful)

Aurige:
Group Decision: Ok

trurlurl:
@nick-knowles Isn't the diagram outdated anyway and would have to be adapted even if we didn't apply any change to the code? 

- AccessModeEnum: `ride` not in the xsd
- AllModesEnum: `anyMode` -> `all`
- `PublicTransportVehicleModeEnum` -> `AllPublicTransportModesEnum`
-- `other` removed, others added: `intercityRail`, `urbanRail`, `taxi`, `anyMode`

How to proceed with the PR, what would be your preferred solution?

Aurige:
NK to review and AM to update accordingly if needed

nick-knowles:
Ride is for an animal ride - horse, mule, camel etc. Might be better as **animalRide**. We should support for personal ojurneys

For modes There is subtle difference between ' all' - e.g.  this station supports the use of all possible  modes to access it ;  versus 'any' -  for this journey you must select one specific mode , any one possibile  mode is allowed.   However it is probably sufficient to just use all. 

it is a moot point as to whether at the MODE level there is just **rail** (or rather heavy rail as opposed to a metro) ,  and any finer distinction such as  between intercityRail and urbanRail should be made at a SUBMODE level, or whether intercityRail and  are actually completely different separate modes in their own right.     It is a bit messy in reallife as  'intercity' may encompass,  TFVhighspeed,  international rail, interregional rail, regional rail,  non TGV intercity etc , and since not all intercity rail strictly requires a reservation. (though all TGV rail does?)  there is no simple criteria to assigne.  I would leave it to the rail experts to say why it is important to make a distinction or not.

nick-knowles:
Also Better to leave taxi as an access mode mode only as it doesnt have a schedule.

trurlurl:
@nick-knowles Having read your comments...

-  I deprecated *intercityRail* and *urbanRail*. They already were deprecated in *AllPublicTransportModesEnumeration*, so we also gain more consistency. 
- I added *animalRide*.
- I didn't change anything about *taxi* since the definition of *AllPublicTransportModesEnumeration* states the following, according to which taxi is a public transport mode: "PUBLIC TRANSPORT MODE: A MODE used for public transport services (e.g., by bus, by tram, by metro, by train, by ferry, by ship, by shared taxi car, by shared car) under either a CONVENTIONAL MODE OF OPERATION or an ALTERNATIVE MODE OF OPERATION (i.e., excluding a PERSONAL MODE OF OPERATION). Note: Prior to TM 6.2 this was named VEHICLE MODE."

Aurige (approved):
I'd like to close this one so we can update the documentation accordingly for 2.0

ue71603 (dismissed):

ue71603 (approved):

ue71603:
@trurlurl @Ulf9 @Aurige @skinkie @duexw : Updated to remove validation problem pls reconfirm

skinkie (approved):

Aurige (approved):

PR #792, opened 2024-10-24
Remove unused chargingEquipmenProfilesInFrame_RelStructure

trurlurl:
`chargingEquipmenProfilesInFrame_RelStructure` has three flaws: a typo (..EquipmenProf..), it should be renamed to *re*changing..., and it is unused. 

Instead of this structure, we have now `vehicleEquipmentProfilesInFram_RelStructure` that does the job of referencing the `RechargingEquipmentProfile`s.


Aurige:
Good catch

Aurige (approved):

ue71603 (approved):

PR #800, opened 2024-11-18
Type of PublicCode

trurlurl:
All `PublicCode` elements should be of type `PublicCodeStructure`, `LinePublicCode` probably as well.

PR #802, opened 2024-11-19
Do the same as #800 but then on master

skinkie:


skinkie:
So no PublicCodeStructure yet. Then only the typo is relevant. I'm closing the bug.

PR #803, opened 2024-11-19
Type of PublicCode

trurlurl:
This is part of PR #800 transposed to master. One remaining occurrence of `PublicCode` needs to be done in next, doesn't exist in master.


skinkie:
We did the same see #802. Does not work because master does not have PublicCodeStructure.

You can submit the typo fix, if you want.

trurlurl:
Haha :-) ..

Issue #399, opened 2022-09-28
NameOfRefClass default


skinkie:
At this moment NameOfRefClass is not defined, and some sort of expected to choose the right one. Having the defaults in place will help deserialisation and resolving.

ue71603:
@skinkie  do you do the PR?

PR #807, opened 2024-12-04
Proposal for #399

skinkie:
In order to futher the restrict the schema to sensible values and relations I would like to introduce two things:
1. Have an **explicit** default for non-generic nameOfRefClass attributes, in the case below, ScheduledStopPointRef has a natural nameOfRefClass being _ScheduledSopPoint_. Now a _parser_ does not have to guess what the default might be. And this could also help with schematron validation, since conditional key-identity-constraints are still _far away_.
2. Further we would like to have a sane list of Elements that _might_ be referenced. From my perspective this should include all objects up to some predefined point. I think that should be something like "every parent, excluding abstract parents, up to and not included DataManagedObject". I would like to prevent siblings to be included.

### DIFF
--- a/xsd/netex_part_1/part1_tacticalPlanning/netex_servicePattern_support.xsd
+++ b/xsd/netex_part_1/part1_tacticalPlanning/netex_servicePattern_support.xsd
@@ -94,6 +94,17 @@ Rail transport, Roads and Road transport
 						<xsd:documentation>Identifier of SCHEDULED STOP POINT.</xsd:documentation>
 					</xsd:annotation>
 				</xsd:attribute>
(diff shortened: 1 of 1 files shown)


ue71603:
@skinkie  does this not result in automatic cycles in the necesseray import of elements?

skinkie:
> @skinkie does this not result in automatic cycles in the necesseray import of elements?

Not really, because the base element is a "xs:string" hence, is a pure text description, not limited to the actual elements defined.

ue71603:
Ok, I see it is only the text "TimingPoint"

ue71603:
Stefan: Help needed. Done for every ref structure

Issue #597, opened 2023-12-13
Write code to update the deprecated examples from PrivateCode to privateCodes/PrivateCode


skinkie:
Helps #594

ue71603:
@skinkie what exactly should be done?

skinkie:
@ue71603 take for example this example;

```xml
<ResponsibilitySet id="ch:1:ResponsibilitySet:BuS_BuS" version="any">
        <Name lang="de">Bus und Service AG</Name>
        <PrivateCode>BuS</PrivateCode>
        <roles> 
                <ResponsibilityRoleAssignment id="ch:1:ResponsibilityRoleAssignment:BuS_BuS:1" version="any">
                        <StakeholderRoleType>entityLegalOwnership</StakeholderRoleType>
                        <ResponsibleOrganisationRef ref="ch:1:Operator:766" version="any"/>
                </ResponsibilityRoleAssignment>
                <ResponsibilityRoleAssignment id="ch:1:ResponsibilityRoleAssignment:BuS_BuS:2" version="any">
                        <StakeholderRoleType>operation</StakeholderRoleType>
                        <ResponsibleOrganisationRef ref="ch:1:Operator:766" version="any"/>
                </ResponsibilityRoleAssignment>
        </roles>
</ResponsibilitySet>
```

This should be changed to:

```xml
<ResponsibilitySet id="ch:1:ResponsibilitySet:BuS_BuS" version="any">
        <privateCodes><PrivateCode>BuS</PrivateCode></privateCodes>
        <Name lang="de">Bus und Service AG</Name>
        <roles> 
                <ResponsibilityRoleAssignment id="ch:1:ResponsibilityRoleAssignment:BuS_BuS:1" version="any">
                        <StakeholderRoleType>entityLegalOwnership</StakeholderRoleType>
                        <ResponsibleOrganisationRef ref="ch:1:Operator:766" version="any"/>
                </ResponsibilityRoleAssignment>
                <ResponsibilityRoleAssignment id="ch:1:ResponsibilityRoleAssignment:BuS_BuS:2" version="any">
                        <StakeholderRoleType>operation</StakeholderRoleType>
                        <ResponsibleOrganisationRef ref="ch:1:Operator:766" version="any"/>
                </ResponsibilityRoleAssignment>
        </roles>
</ResponsibilitySet>
```

The difficult part is that privateCodes must be moved before the Name here.

ue71603:
so just do some example correction work?
[...]

PR #811, opened 2024-12-04
privateCodes implemented (partically) in all functional elements the first is now in the privateCodes part.

ue71603:
use privateCodes and no longer the depreciated PrivateCode: 
https://github.com/NeTEx-CEN/NeTEx/issues/597

First version: Should it be done this way for all examples?


ue71603:
It never is. It is always in a different position, but I think the number is not too big

ue71603:
and having "most". would also be enough in my view.

Aurige:
That's fine to me.
That's also something that we can do in a longer run: fixing the examples does not require a CEN NWI or a document update ... so we can directly update master whenever we find a new one to be fixed.

trurlurl (approved):

Aurige (approved):

PR #846, opened 2025-04-07
WIP: Updating Quay/StoPlace

JohanEntur:
Discussed in https://3.basecamp.com/3256016/buckets/2570434/messages/8335437909.
I wanted this to be discussed first, but since we are running out of meetings, and this doesnt come up, I'll just post this right away.

Again, thanks to my colleague for composing the actual PR.

### DIFF
--- a/xsd/netex_framework/netex_reusableComponents/netex_mode_support.xsd
+++ b/xsd/netex_framework/netex_reusableComponents/netex_mode_support.xsd
@@ -436,4 +436,17 @@ Rail transport, Roads and Road transport
 			<xsd:enumeration value="other"/>
 		</xsd:restriction>
 	</xsd:simpleType>
+	<xsd:simpleType name="StopPlaceCategoryEnumeration">
+		<xsd:annotation>
+			<xsd:documentation>Category of STOP PLACE.</xsd:documentation>
+		</xsd:annotation>
+		<xsd:restriction base="xsd:string">
+			<xsd:enumeration value="airport"/>
+			<xsd:enumeration value="terminal"/>
+			<xsd:enumeration value="station"/>
+			<xsd:enumeration value="hub"/>
+			<xsd:enumeration value="halt"/>
+			<xsd:enumeration value="unregulated"/>
+		</xsd:restriction>
+	</xsd:simpleType>
 </xsd:schema>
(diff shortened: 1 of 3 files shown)


skinkie:
Is QuayLayoutTypeEnumeration in use?

Can the different types have proper documentation?

ue71603:
with proper documentation Stefan means:
```
			<xsd:enumeration value="lane">
					<xsd:annotation>
						<xsd:documentation>What does this really mean?</xsd:documentation>
					</xsd:annotation>
			</xsd:enumeration>

```
Also, always add such a thing to one of the appropriate example file (each element should be used at least in one example.
Here e.g. one in https://github.com/NeTEx-CEN/NeTEx/tree/next/examples/functions/site

ue71603:
Otherwise looks good.

xsd/netex_framework/netex_reusableComponents/netex_mode_support.xsd
```diff
@@ -436,4 +436,17 @@ Rail transport, Roads and Road transport
 			<xsd:enumeration value="other"/>
 		</xsd:restriction>
 	</xsd:simpleType>
+	<xsd:simpleType name="StopPlaceCategoryEnumeration">
```
trurlurl, line 439: What's the difference between "Category" and "Type" - i.e., why has the name *StopPlace**Category**Enumeration* been chosen instead of *StopPlace**Type**Enumeration* (similar to the existing *Stop**Type**Enumeration* and other type names)?

Aurige:
I initially thought that this would be for next release (not for v2), since it requires quite a lot of work:

- articulation with existing enum: we will probably need to deprecate some values, and articulation between StopTypeEnumeration and StopPlaceCategoryEnumeration (may be chose a better name as pointed by @trurlurl ), also articulation between busStopWithinRoadwayBoarding or tramStopWithinRoadwayBoarding with the new QuayLayoutTypeEnumeration, etc.
- for now the StopPlaceCategory is only in the StopPlaceView which is king of limited
- it will also most probably require some discussion with the Transmodel group
- etc.

So I would rather target it to a milestone after 2.0

ue71603:
@Aurige ok for me either way. moved to 2.1 for the time being

TuThoThai:
Changing the base to `v2.1-wip`

PR #852, opened 2025-04-10
PrintColourValueType

trurlurl:
The changes reflect my understanding that everything should be using print colours and that print colours are not limited to CMYK.


skinkie:
Difficult, difficult... I think you want to have some sensible RGB value for onscreen production, but that will open another can of worms.

skinkie (dismissed):

ue71603 (approved):

Aurige (approved):

PR #860, opened 2025-04-21
Remove all order attributes in favor of XML ordering.

### LINKED ISSUE #810: Remove all order attributes which are informative
In #560 we have identified that the StopPointInJourneyPatternRef in virtually all examples ends up _incomplete_ and is therefore hidden in validation because the referee does not provide all required field. From XML Schema perspective, [the explicit order is optional](https://gist.github.com/scottoffen/58dd16d0b8408b713140), since the sequence already requires ordered processing. I do not dispute that it is nice to have a numeric value specifying order, but since the id+version element on its own must be unique anyway I would say that removing the order element all together for sequences makes a lot more sense. Obviously, this is XML specific. Any other implementation should take into account that the sequence is in fact an OrderedList and not just a List.

skinkie:
Fix #810

This has uncovered quite some issues in the example files, including referencing issues where some referencing entities have not been correctly been assigned. A fix for those should be a separate pull request, build from this one.


TuThoThai (approved):
Looks all good for me on the XSD parts. Did not look at the examples as I trust you on them @skinkie

skinkie:
I want votes! All examples fixed!

TuThoThai (approved):

Aurige (approved):

Issue #777, opened 2024-09-08
Centroids without  Location and Coordinates should no longer be allowed. I also think that WSG84 always must be present


ue71603:

skinkie:
The issue with Centroid, it is allowed to to be _nothing_, so we must make LocationStructure mandatory:

```xsd
<xsd:complexType name="SimplePoint_VersionStructure" abstract="false">
        <xsd:annotation>
                <xsd:documentation>Type for a Simple POINT.</xsd:documentation>
        </xsd:annotation>
        <xsd:complexContent>
                <xsd:extension base="EntityInVersionStructure">
                        <xsd:sequence>
                                <xsd:element name="Name" type="MultilingualString" minOccurs="0">
                                        <xsd:annotation>
                                                <xsd:documentation>Name of POINT.</xsd:documentation>
                                        </xsd:annotation>
                                </xsd:element>
                                <xsd:element name="Location" type="LocationStructure" minOccurs="0">
                                        <xsd:annotation>
                                                <xsd:documentation>The position of a POINT with a reference to a given LOCATING SYSTEM (e. g. coordinates).</xsd:documentation>
                                        </xsd:annotation>
                                </xsd:element>
                        </xsd:sequence>
                </xsd:extension>
        </xsd:complexContent>
</xsd:complexType>
```

...and then, the location structure specifies Latitude/Longitude and gml:Pos, but we don't specify that either of them is mandatory. And my guess is, this is done because we don't want a choice here (both must be able to be provided). And we don't have any way to assert that any of them is provided.

```xsd
<xsd:complexType name="LocationStructure">
        <xsd:annotation>
                <xsd:documentation>Type for geospatial Position of a point. May be expressed in concrete WGS 84 Coordinates or any gml compatible point coordinates format.</xsd:documentation>
        </xsd:annotation>
        <xsd:sequence>
                <xsd:sequence>
                        <xsd:sequence minOccurs="0">
                                <xsd:annotation>
                                        <xsd:documentation>WGS84</xsd:documentation>
                                </xsd:annotation>
                                <xsd:element name="Longitude" type="LongitudeType">
                                        <xsd:annotation>
                                                <xsd:documentation>Longitude from Greenwich Meridian. -180 (East) to +180 (West). Decimal degrees. e.g. 2.356</xsd:documentation>
                                        </xsd:annotation>
                                </xsd:element>
                                <xsd:element name="Latitude" type="LatitudeType">
                                        <xsd:annotation>
[...]

PR #869, opened 2025-04-21
Improve the likelihood for Longitude and Latitude and having a coordinate for each Location

skinkie:
As discussed in #777

The result of the discussion is that Longitude/Latitude should be set.. In DECK PLAN we have relative coordinates that need to be done with gml:pos. The documentation is adapted.


skinkie:
It seems that Deckplan invented a relative grid. Decide what to do with that if WGS84 would be mandatory.

```
<DeckPathJunction version="any" id="rc:DPJ-slw_deck_1@forward_middle">
        <!-- locationgiven as offset from front left boundingbox of parent space or use projection  -->
        <Name>mid point at front fo carriage</Name>
        <Location>
                <gml:pos srsName="xy_in_Metres">3 2</gml:pos>
        </Location>
        <PublicUse>all</PublicUse>
        <PassengerSpaceRef version="any" ref="rc:PS-slw_deck_1@corridor"/>
</DeckPathJunction>
```

trurlurl (dismissed):

Aurige:
That's more generic than DeckPlan (which is obviously not geographic): there is a specific need for any pathlink on a schematic maps that may not be geographic (that's an Indoor Mapping & Indoor Cartography issue). That's most probably useful for EPIAP too (a lot of underground maps and station maps are not geographic... there is also different habits: France tends to uses geographic maps, where UK often prefers schematic, non geographic, ones).

skinkie:
That would either justify a different type, or not making WGS84 mandatory (or worse: the choice option).

ue71603:
@Aurige My problems:
* The calculation of projectiond cannot be the duty of the data receiver in an European context.
* I don't want any geo localisaed object withouth geolocalisation.

I may accept that there are relative coordinates like deckplan. For indoor routing I think this ALL has to work with GPS as a baseline and therefore needs appropriate coordinates.

Aurige:
I wanted to propose something like this (to make one of the location coordinate mandatory), but for some reason, it triggers validation issues (maybe an XMLSpy bug)
image

	<xsd:complexType name="LocationStructure">
		<xsd:annotation>
			<xsd:documentation>Type for geospatial Position of a point. May be expressed in concrete WGS 84 Coordinates or any gml compatible point coordinates format.</xsd:documentation>
		</xsd:annotation>
		<xsd:sequence>
			<xsd:choice>
				<xsd:sequence>
					<xsd:group ref="WgsGroup"/>
					<xsd:element ref="gml:pos"/>
				</xsd:sequence>
				<xsd:sequence>
					<xsd:group ref="WgsGroup"/>
				</xsd:sequence>
				<xsd:sequence>
					<xsd:element ref="gml:pos"/>
				</xsd:sequence>
			</xsd:choice>
			<xsd:element name="Precision" type="DistanceType" minOccurs="0">
				<xsd:annotation>
					<xsd:documentation>Precision for point measurement. In meters.</xsd:documentation>
				</xsd:annotation>
			</xsd:element>
		</xsd:sequence>
		<xsd:attribute name="id" type="xsd:normalizedString">
			<xsd:annotation>
				<xsd:documentation>Identifier of point.</xsd:documentation>
			</xsd:annotation>
		</xsd:attribute>
		<xsd:attribute name="srsName" type="SrsNameType">
			<xsd:annotation>
				<xsd:documentation>identifier of data Reference system for geocodes if point is specified as gml compatible Coordinates. A gml value. If not specified taken from system configuration.</xsd:documentation>
			</xsd:annotation>
		</xsd:attribute>
	</xsd:complexType>

ue71603:
@Aurige You didn't put in the problem that occurs. I think it can't decide the path, does it?

ue71603:
I fixed it already and it should now accept what we want with the last two commits. Is there still a problem?
The ambiguous problem is: if there is e.g  the wgsgroup. XMLSpy does not know from which branch of the schema this comes.

from the  wgsgroup or gml:pos or the wgsgroup_only

It does not like that.

Aurige (approved):

skinkie:
@ue71603 I really dislike how this is rendered into an api-interface. Have you checked that?

ue71603:
> @ue71603 I really dislike how this is rendered into an api-interface. Have you checked that?
No idea what you are talking about.

Aurige:
Don't merge (due to code generation issue) ... or find a better solution
Add comment about deprecating WGS
Add a comment explicitly stating that a position is MANDATORY (GML or WGS for now) 
+ Doc updates

PR #881, opened 2025-04-26
Fix correct anpr enumeration value on machine readable

nick-knowles:
FIX correct anpr  enumeration value on MachineReadable


ue71603:

PR #884, opened 2025-04-28
Tm Alignment  Add SPOT ALLOCATION Elements and updated examples to show use.

nick-knowles:
Add to SalesTransaction models in Part3 Fares.  Update an Example to cover


ue71603:

PR #886, opened 2025-04-28
TM ALignment DECK PLANS Add type of deck  entrance usage to DECK ENTRANCE USAGE 

nick-knowles:
And an open TypeOfDEckENtrance  to DECK ENTRANCE USAGE


ue71603:

PR #887, opened 2025-04-29
Tmlign  Add max poer to parking 

nick-knowles:
This allow us s to use PARKING as a PARKING STATION


ue71603:

PR #891, opened 2025-05-01
Fixes to Examples APrl 2025 - Change to order constraint etc

nick-knowles:
The following changes are needed to get a clean validation  with XML SP

Most are to adapt fro the change to the key constraint to remove order from the key

Teh AXML SPY project spp file sis also updated to add / remove some other examples


ue71603:
will be done in #893

PR #813, opened 2024-12-11
Seating Plan  Seat Context changes

nick-knowles:
Changes from Netex meeting
(a) Remove hasTray  
(b) Add blankRowEndNoAisle to seat context
(c) Update example

PR #909, opened 2025-05-26
SeatContext proposal

### LINKED ISSUE #701: SeatContext should be boolean or a list of values
In the pull request SeatContext was introduced as an enumeration, instead of boolean values or list of values. I don't think the SeatContext is an improvement, and should be reverted.

TuThoThai:
Addresses both issue #701 and group decision made on 18 March 2025 in PR #813

### DIFF
--- a/examples/functions/deckPlans/DeckPlans-Example_train_simple.xml
+++ b/examples/functions/deckPlans/DeckPlans-Example_train_simple.xml
@@ -578,9 +578,10 @@
 													</actualVehicleEquipments>
 													<SpotRowRef version="any" ref="rc:SR-bfc_deck_1@row_01"/>
 													<SpotColumnRef version="any" ref="rc:SC-bfc_deck_1@col_A"/>
-													<SeatContext>byWindowNoAisle</SeatContext>
 													<TableType>none</TableType>
 													<HasPower>true</HasPower>
+													<ByWindow>true</ByWindow>
+													<ByAisle>false</ByAisle>
 												</PassengerSpot>
 												<PassengerSpot version="any" id="rc:PSp-bfc_deck_1@seating_area@row_01_col_B">
 													<Label>18</Label>
@@ -592,9 +593,10 @@
 													</actualVehicleEquipments>
 													<SpotRowRef version="any" ref="rc:SR-bfc_deck_1@row_01"/>
 													<SpotColumnRef version="any" ref="rc:SC-bfc_deck_1@col_B"/>
-													<SeatContext>byAisleNoWindow</SeatContext>
 													<TableType>none</TableType>
 													<HasPower>true</HasPower>
+													<ByWindow>false</ByWindow>
+													<ByAisle>true</ByAisle>
 												</PassengerSpot>
 												<PassengerSpot version="any" id="rc:PSp-bfc_deck_1@seating_area@row_01_col_C">
 													<Label>14</Label>
@@ -606,9 +608,10 @@
 													</actualVehicleEquipments>
 													<SpotRowRef version="any" ref="rc:SR-bfc_deck_1@row_01"/>
 													<SpotColumnRef version="any" ref="rc:SC-bfc_deck_1@col_C"/>
-													<SeatContext>byAisleNoWindow</SeatContext>
 													<TableType>none</TableType>
-													<HasPower>true</HasPower>
+													<HasPower>true</HasPower>													
+													<ByWindow>false</ByWindow>
+													<ByAisle>true</ByAisle>
 												</PassengerSpot>
 												<PassengerSpot version="any" id="rc:PSp-bfc_deck_1@seating_area@row_01_col_D">
 													<Label>16</Label>
@@ -620,9 +623,10 @@
 													</actualVehicleEquipments>
 													<SpotRowRef version="any" ref="rc:SR-bfc_deck_1@row_01"/>
 													<SpotColumnRef version="any" ref="rc:SC-bfc_deck_1@col_D"/>
-													<SeatContext>byWindowNoAisle</SeatContext>
 													<TableType>none</TableType>
-													<HasPower>true</HasPower>
+													<HasPower>true</HasPower>													
+													<ByWindow>true</ByWindow>
+													<ByAisle>false</ByAisle>
 												</PassengerSpot>
 												<!-- Row 02 for 21-27 (forward facing)-->
 												<PassengerSpot version="any" id="rc:PSp-bfc_deck_1@seating_area@row_02_col_A">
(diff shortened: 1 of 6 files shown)


skinkie (dismissed):

Aurige (dismissed):

ue71603 (dismissed):

xsd/netex_framework/netex_reusableComponents/netex_seatingPlan_version.xsd
```diff
@@ -415,6 +410,41 @@
...
+				<xsd:annotation>
+					<xsd:documentation>Whether PASSENGER SPOT is in the front row.</xsd:documentation>
+				</xsd:annotation>
+			</xsd:element>
+			<xsd:element name="IsInEndRow" type="xsd:boolean" minOccurs="0">
+				<xsd:annotation>
+					<xsd:documentation>Whether PASSENGER SPOT is in the end row, next to a wall.</xsd:documentation>
+				</xsd:annotation>
+			</xsd:element>
+			<xsd:element name="HasWindowinFront" type="xsd:boolean" minOccurs="0">
```
skinkie, line 438:
```suggestion
			<xsd:element name="HasWindowInFront" type="xsd:boolean" minOccurs="0">
```

xsd/netex_framework/netex_reusableComponents/netex_seatingPlan_version.xsd
```diff
@@ -415,6 +410,41 @@
...
+				</xsd:annotation>
+			</xsd:element>
+			<xsd:element name="ByAisle" type="xsd:boolean" minOccurs="0">
+				<xsd:annotation>
+					<xsd:documentation>Whether PASSENGER SPOT is next to an aisle.</xsd:documentation>
+				</xsd:annotation>
+			</xsd:element>
+			<xsd:element name="IsInMiddle" type="xsd:boolean" minOccurs="0">
+				<xsd:annotation>
+					<xsd:documentation>Whether PASSENGER SPOT is in the middle, i.e. not next to aisle or  a window with seat on either side.</xsd:documentation>
```
skinkie, line 425:
```suggestion
					<xsd:documentation>Whether PASSENGER SPOT is in the middle, i.e. not next to aisle or a window with seat on either side.</xsd:documentation>
```

skinkie (dismissed):

Robbendebiene:
Just wanted to add some thoughts:
- Some values describe surrounding things (`ByWindow`, `ByAisle`) while others describe the position (`IsInMiddle`, `IsInFrontRow`) and orientation (`HasWindowInFront`) of the seat.
- I think changing `IsInMiddle` to `BetweenSeats` would be more clear.
- How is front and end defined for `IsInFrontRow` and `IsInEndRow`? Isn't the interesting part the surrounding of the seat? e.g. `ByWall`.
- `HasAisleInFront` doesn't seem to add much value to `ByAisle`.

TuThoThai:
> Also what is the reason to go for several boolean parameters instead of an enum list as already suggested by @JohanEntur and @skinkie in the related issue? For me the main benefit of an enum list would be that the values are logically grouped together under an umbrella term like SeatContext. If they are separated booleans then they lose the context/use case for which the values should be used.

@Robbendebiene, there was a long discussion in March between the enum and the list of booleans. The consensus was that:
- yes, the enum makes it easier since everything is under one umbrella
- but, the enum lacks clarify and some values could have led to very different interpretations and without certainty that they cover all possible cases
So, the group agreed to have a series of booleans that also be combined.  

>  How is front and end defined for `IsInFrontRow` and `IsInEndRow`? Isn't the interesting part the surrounding of the seat? e.g. `ByWall`.

For me, on that one, the front is defined as the closest to the driver cabin / carriage when looking at the direction in which the vehicle is going. Or at least it is how I wanted to describe it in the documentation.
`IsInFrontRow` != `ByWall` if you are on a bus shuttle where the front row is not necessarily near a wall / glass panel but next to the driver.

> `HasAisleInFront` doesn't seem to add much value compared to `ByAisle`.

The main difference I see is in how you get in / out of the seat, and it was useful for me to describe the middle seat of the end row in the bus example. But, I am happy to drop it if it is the general consensus in the group. 

> I think changing `IsInMiddle` to `BetweenSeats` would be more clear.

I have no strong opinion, here. Either way is fine for me. So, depending on what @skinkie, @Aurige, @JohanEntur prefer, I can make the change.

skinkie:
I would then say IsBetweenSeats to make clear it is a boolean.

Robbendebiene:
> > I'm still not sure about the arguments against the enum.
> 
> Because it is far more difficult to parse.

@skinkie  Out of curiosity, what makes it difficult to parse? Since basically every facility is a `SomeEnumListOfEnumerations` are they hard to parse as well? I can see that boolean properties translate easily into boolean variables of a class but shouldn't a `SomeEnumListOfEnumerations` also easily translate into a single variable of type `List<SomeEnum>`?

Robbendebiene:
Last suggestion from my side 😇
If `HasWindowInFront` and `HasAisleInFront` should be kept then perhaps turning them into the same format as the other booleans (`IsABC`) makes sense to "artificially" group them together?

- `HasWindowInFront` -> `IsFacingWindow` or `IsBeforeWindow`
- `HasAisleInFront` -> `IsFacingAisle` or `IsBeforeAisle`

skinkie:
> @skinkie Out of curiosity, what makes it difficult to parse? Since basically every facility is a `SomeEnumListOfEnumerations` are they hard to parse as well? I can see that boolean properties translate easily into boolean variables of a class but shouldn't a `SomeEnumListOfEnumerations` also easily translate into a single variable of type `List<SomeEnum>`?

The biggest problem is that the changes made by Nick basically combined variants (permutation) in single instances, this is what we are reverting back to direct class attributes. he SomeEnumeListOfEnumerations leads to complex queries for OR and AND operators, not directly available without explict parsing and handeling of that list with IN like queries, where the checks have to be explicitly modelled based on the _enumeration_.

TuThoThai:
> Last suggestion from my side 😇 If `HasWindowInFront` and `HasAisleInFront` should be kept then perhaps turning them into the same format as the other booleans (`IsABC`) makes sense to "artificially" group them together?
> 
>     * `HasWindowInFront` -> `IsFacingWindow` or `IsBeforeWindow`
> 
>     * `HasAisleInFront` -> `IsFacingAisle` or `IsBeforeAisle`

Works for me. I like`IsFacingWindow` and `IsFacingAisle` better as they sound clearer to me. Updates on the PR coming!

skinkie (dismissed):

examples/functions/deckPlans/DeckPlans-Example_train_simple.xml
```diff
@@ -777,9 +791,10 @@
 													</actualVehicleEquipments>
 													<SpotRowRef version="any" ref="rc:SR-bfc_deck_1@row_04"/>
 													<SpotColumnRef version="any" ref="rc:SC-bfc_deck_1@col_C"/>
-													<SeatContext>byAisleNoWindow</SeatContext>
 													<TableType>none</TableType>
 													<HasPower>true</HasPower>
+													<IsByWindow>false</IsByWindow>
+													<IsByAisle>true</IsyAisle>
```
skinkie, line 797:
```suggestion
													<IsByAisle>true</IsByAisle>
```

skinkie:
@TuThoThai some more validation errors...

TuThoThai:
> @TuThoThai some more validation errors...

@skinkie, apologies... my eyes were clearly not working properly! Thanks for fixing the errors!

skinkie (approved):

ue71603 (approved):

PR #914, opened 2025-05-30
fixing From/To JourneyRefStructure

ue71603:
Its From/To JourneyRef so the type is Journey and not VehicleJourney. 
This is also necessary, because VehicleJourney and ServiceJourney are both derived from Journey, but not one form the other.

### DIFF
--- a/xsd/netex_part_2/part2_journeyTimes/netex_interchange_version.xsd
+++ b/xsd/netex_part_2/part2_journeyTimes/netex_interchange_version.xsd
@@ -159,26 +159,52 @@ Rail transport, Roads and Road transport
 					<xsd:documentation>SCHEDULED STOP POINT at which JOURNEY MEETING takes place.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
-			<xsd:element name="FromJourneyRef" type="JourneyRefStructure">
-				<xsd:annotation>
-					<xsd:documentation>JOURNEY that feeds JOURNEY MEETING.</xsd:documentation>
-				</xsd:annotation>
-			</xsd:element>
-			<xsd:element name="ToJourneyRef" type="JourneyRefStructure">
-				<xsd:annotation>
-					<xsd:documentation>JOURNEY that distributes from JOURNEY MEETING.</xsd:documentation>
-				</xsd:annotation>
-			</xsd:element>
-			<xsd:element name="FromPointInJourneyPatternRef" type="PointInJourneyPatternRefStructure" minOccurs="0">
-				<xsd:annotation>
-					<xsd:documentation>POINT IN JOURNEY PATTERN ofr  feeder  journey JOURNEY PATTERN.</xsd:documentation>
-				</xsd:annotation>
-			</xsd:element>
-			<xsd:element name="ToPointInJourneyPatternRef" type="PointInJourneyPatternRefStructure" minOccurs="0">
-				<xsd:annotation>
-					<xsd:documentation>POINT IN JOURNEY PATTERN ofr  distributorjourney JOURNEY PATTERN.</xsd:documentation>
-				</xsd:annotation>
-			</xsd:element>
+			<xsd:choice>
+				<xsd:sequence>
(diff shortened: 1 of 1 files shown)


xsd/netex_part_2/part2_journeyTimes/netex_interchange_version.xsd
```diff
@@ -159,14 +159,14 @@ Rail transport, Roads and Road transport
...
 				<xsd:annotation>
-					<xsd:documentation>VEHICLE JOURNEY that feeds JOURNEY MEETING.</xsd:documentation>
+					<xsd:documentation>JOURNEY that feeds JOURNEY MEETING.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
-			<xsd:element name="ToJourneyRef" type="VehicleJourneyRefStructure">
+			<xsd:element name="ToJourneyRef" type="JourneyRefStructure">
 				<xsd:annotation>
-					<xsd:documentation>VEHICLE JOURNEY  that distributes from JOURNEY MEETING.</xsd:documentation>
+					<xsd:documentation>JOURNEY  that distributes from JOURNEY MEETING.</xsd:documentation>
```
skinkie, line 169:
```suggestion
					<xsd:documentation>JOURNEY that distributes from JOURNEY MEETING.</xsd:documentation>
```

skinkie (dismissed):

trurlurl (dismissed):

xsd/netex_part_2/part2_journeyTimes/netex_interchange_version.xsd
```diff
@@ -159,14 +159,14 @@ Rail transport, Roads and Road transport
 					<xsd:documentation>SCHEDULED STOP POINT at which JOURNEY MEETING takes place.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
-			<xsd:element name="FromJourneyRef" type="VehicleJourneyRefStructure">
+			<xsd:element name="FromJourneyRef" type="JourneyRefStructure">
```
Aurige, line 162: I don't understand here : the Transmodel definition for JOURNEY MEETING is "A time constraint for one or several **SERVICE JOURNEYs** fixing interchanges between them and/or an external event (e.g. arrival or departure of a feeder line, opening time of the theatre, etc.)."

Aurige: That's right, but that would require a deprecation in 2.0

ue71603: and it should be a SERVICE JOURNEY MEETING then. And it is still not working correctly, because the whole hierarchy is messed up, because ServiceJourney is not derived from VerhicleJourney but from Journey. So by going through Journey it works.
Also: switching from quay  3 to 4 before continuing is not a ServiceJourney in any system. With Journey it still holds.

So No, I will not accept Stefan's changes.

ue71603: because this is not a feature fix, this is a bug fix name and what the hierarchy does are aligned afterwards. Your change you can do for 3.0

skinkie: @ue71603 reverted, your suggestion for the type of ConnectingJourneyRef

ue71603:
still not happy, but it solves the problem of Christophe and you. so let's do this.

skinkie:
What are you not happy about, this was your proposal? :-)

ue71603:
I know. but you see that I reverted the reversion and we do the depreciation as you and Christophe discussed. Christophe says it should be a ServiceJourney and I am not that sure.. and the problem if a ServiceJourney is a specialed VehicleJourney is not solved. But it is an improvement and we are out of time.

Aurige:
The fact that is is a ServiceJourney is from Transmodel ... if we want to change it, it is another level of discussion
I'm also happy with the improvement from that PR, even if additional discussion may be needed

Aurige (approved):

skinkie (approved):

Issue #562, opened 2023-12-05
TransportAdministrativeZone has a list Description? 


skinkie:
I think it boils down to `AbstractGroupMember_VersionedChildStructure`. Can some of our XmlSpy overlords checks if this hunch is correct?

```
        <xsd:complexType name="`" abstract="true">
                <xsd:annotation>
                        <xsd:documentation>Type for a GROUP OF ENTITY MEMBERs. Subclass this for specific member types.</xsd:documentation>
                </xsd:annotation>
                <xsd:complexContent>
                        <xsd:extension base="VersionedChildStructure">
                                <xsd:sequence>
                                        <xsd:element name="Description" type="MultilingualString" minOccurs="0"/>
                                </xsd:sequence>
                                <xsd:attribute name="order" type="xsd:positiveInteger">
                                        <xsd:annotation>
                                                <xsd:documentation>Order of member within parent group.</xsd:documentation>
                                        </xsd:annotation>
                                </xsd:attribute>
                        </xsd:extension>
                </xsd:complexContent>
        </xsd:complexType>
```

ue71603:
@skinkie  I am not that shure what you want. Some diagrams. Do they help?
![image](https://github.com/NeTEx-CEN/NeTEx/assets/24227470/a782818f-c217-4820-9f83-63c327635d76)
![image](https://github.com/NeTEx-CEN/NeTEx/assets/24227470/8ddb28d7-21f8-4a2d-95c0-2f33f6d163ca)

Or what exactly should I check.

skinkie:
@ue71603 "Description" if Description here suggest there can be multiple elements, the XML Schema has a bug, otherwise the binding has a bug.

ue71603:
![image](https://github.com/NeTEx-CEN/NeTEx/assets/24227470/ebebed53-eb74-476d-b431-36a476d00d30)

skinkie:
As non XML spy user, what do I see? Multiple or single?

ue71603:
The is 0 -1 Description...

nick-knowles:
The hierarchy is 

DataManagedObject
   GroupOfEntities
      GroupOfPoints
         Zone
           AdministrativeZone
            TransportAdministrativeZone

I donyt see ant reference to version child  - what am I missing?

ue71603:
@skinkie  still something to do here?

skinkie:
@ue71603 these issues still exist in next:
```
[...]

PR #917, opened 2025-06-08
Remove redundant Description

skinkie:
Fix for #562


ue71603 (approved):

Aurige (commented):
removing IsAllowed is ok (it's new), but Description should have been deprecated

Issue #405, opened 2023-01-08
Renaming the subsitution group not to be alike with a class (ServiceAccessRight)


skinkie:
I would like to have a discussion on things that are either abstracts or subsitution groups currently suffixed with "_". This causes code generation to do funky things. Is there any way we could just rename them to something so it becomes clear they are not to be used as element ever? This is at least the case for ServiceAccessRight.

https://github.com/NeTEx-CEN/NeTEx/blob/master/xsd/netex_part_3/part3_fares/netex_fareProduct_version.xsd#L145

Suggestion to take a look at:

assignment_version_structure
bounding_box_structure
empty_type
extensions
extensions_structure
group_of_entities_ref_structure
line_shape_structure
location_structure
point
point_on_link_ref_structure
series_constraint_ref_structure
service_access_right
srs_name
travel_specification

Aurige:
Objects with a trailing "_" like **ServiceAccessRight_** should probably be abstract 
A trailing "_" also mean that they are not to be used as element ever
But all abstract object don't have a "_", we could have a convention to add it ... I don't think that this would break anything

Aurige:
Discussed: Replace he trailing  "_" with "_Dummy" ... to be done at the end of the revision process by @skinkie

skinkie:
There are currently two situations where I think the extra dummy does not make sense. In both cases they were extended to facilitate _order_. But what was the reason not to add order as an extension, instead of a restriction?

```
        <xsd:complexType name="PointOnLinkRefStructure_">
                <xsd:annotation>
                        <xsd:documentation>Type for a reference to a POINT ON LINK.</xsd:documentation>
                </xsd:annotation>
                <xsd:simpleContent>
                        <xsd:extension base="PointRefStructure">
                                <xsd:attribute name="order" type="xsd:positiveInteger" use="optional">
                                        <xsd:annotation>
                                                <xsd:documentation>Order of point on link. -v2.0</xsd:documentation>
                                        </xsd:annotation>
                                </xsd:attribute>
                        </xsd:extension>
                </xsd:simpleContent>
        </xsd:complexType>
        <xsd:complexType name="PointOnLinkRefStructure">
                <xsd:annotation>
                        <xsd:documentation>Type for a reference to a POINT ON LINK.</xsd:documentation>
                </xsd:annotation>
                <xsd:simpleContent>
                        <xsd:restriction base="PointOnLinkRefStructure_">
[...]

PR #918, opened 2025-06-08
Both complexTypes are in fact abstract.

skinkie:
This pull request aims to resolve the task presented in #405 

First it solves two complexTypes that should have been made abstract. The following will do the renaming.

### DIFF
--- a/xsd/netex_framework/netex_genericFramework/netex_pointAndLink_support.xsd
+++ b/xsd/netex_framework/netex_genericFramework/netex_pointAndLink_support.xsd
@@ -183,7 +183,7 @@ Rail transport, Roads and Road transport
 			<xsd:documentation>Reference to a POINT ON LINK.</xsd:documentation>
 		</xsd:annotation>
 	</xsd:element>
-	<xsd:complexType name="PointOnLinkRefStructure_">
+	<xsd:complexType name="PointOnLinkRefStructure_" abstract="true">
 		<xsd:annotation>
 			<xsd:documentation>Type for a reference to a POINT ON LINK.</xsd:documentation>
 		</xsd:annotation>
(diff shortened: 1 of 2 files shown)


skinkie:
@Aurige I noticed that some of the "element ref" variants actually refer to an "dummy type". I guess this might be a bug in itself.

skinkie:
I guess the real conversation should be:
 1. what should be the suffix, should it be called _DummyType or would _Abstract cover the case better?
 2. is it correct that the _DummyType must always be abstract="true" but abstract="true" should not have to be a _DummyType?
 3. should we suffix the true abstract elements (for whatever reason) with a suffix too?
 4. how is it possible that DummyTypes are used as element ref="blabla_DummyType"?

ue71603 (approved):
Pls make sure that this does not result in all documentation needing a do over.

@trurlurl

skinkie:
> Pls make sure that this does not result in all documentation needing a do over.
> 
> @trurlurl

If it does not chang elementnames, it should never do so.

trurlurl (approved):

Aurige (approved):

nick-knowles:
>I guess the real conversation should be:
>what should be the suffix, should it be called _DummyType or would _Abstract cover the case better?
> is it correct that the _DummyType must always be abstract="true" but abstract="true" should not have to be a _DummyType?
+Yes dummy type should always be abstract,  and _Abstract is okay and probably better than Dummy.  
>should we suffix the true abstract elements (for whatever reason) with a suffix too?
No. In many cases (the normal case)  the  simple abstract type  can be used  as the substitution class on subtypes without introducing a dummy type

I think we want to specifically  mark the subset of cases where  we have had to introduce a  dummy  type  in order to make the XML substitution mechanism work.   It may be at some future point we can revise thiso r resolve what the XML spy issue is.

>how is it possible that DummyTypes are used as element ref="blabla_DummyType"? 
Sounds odd - should review individual cases. Where does this happen

Issue #921, opened 2025-06-13
Should SingleJourney relate to VehicleTypeRef and not VehicleRef?


TuThoThai:
NeTEx part 5

I think there is a bug in the [XSD of NeTEx Part 5](https://github.com/NeTEx-CEN/NeTEx/blob/25ff49bc80646a74f8c187eb4051f041d1918ccc/xsd/netex_part_5/part5_sj/netex_nm_singleJourneyService_version.xsd#L156) for `SingleJourney`:

```xml
	<xsd:group name="SingleJourneyReferencesGroup">
		<xsd:annotation>
			<xsd:documentation>Reference Elements for SINGLE JOURNEY. </xsd:documentation>
		</xsd:annotation>
		<xsd:sequence>
			<xsd:element ref="CommonVehicleServiceRef" minOccurs="0"/>
			<xsd:element ref="VehicleRef" minOccurs="0"/>
			<xsd:element ref="SingleJourneyPathRef" minOccurs="0"/>
		</xsd:sequence>
	</xsd:group>
```

I think, it should read `VehicleTyepRef` instead of `VehicleRef` which would translate in:

```xml
	<xsd:group name="SingleJourneyReferencesGroup">
		<xsd:annotation>
			<xsd:documentation>Reference Elements for SINGLE JOURNEY. </xsd:documentation>
		</xsd:annotation>
		<xsd:sequence>
			<xsd:element ref="CommonVehicleServiceRef" minOccurs="0"/>
			<xsd:element ref="VehicleTypeRef" minOccurs="0"/>
			<xsd:element ref="SingleJourneyPathRef" minOccurs="0"/>
		</xsd:sequence>
	</xsd:group>
```

But, because we have decided to freeze the XSD, I can work on the PR for 2.1

skinkie:
If it is a bug, it should still be fixed in 2.0.

TuThoThai:
@Aurige and/or @ue71603 : could you confirm if it is a bug? If yes, happy to correct it! 😄

skinkie:
I do have a question. If it is a single journey and one knows what it is executed with. VehicleRef is correct right?

TuThoThai:
That is the question I had too. And because I am not so familiar with Part 5, I wasn't sure.
The only hint I had that it might be a bug is that the document refers to Vehicle Type and the XSD was marked as having an error in XMLSpy. 🤔

ue71603:
[...]

PR #924, opened 2025-06-21
Fix netex_nm_singleJourneyService_version.xsd

TuThoThai:
Missing inclusion of `vehicle_support.xsd`

See #921 for further details

### DIFF
--- a/xsd/netex_part_5/part5_sj/netex_nm_singleJourneyService_version.xsd
+++ b/xsd/netex_part_5/part5_sj/netex_nm_singleJourneyService_version.xsd
@@ -3,6 +3,7 @@
 	<xsd:include schemaLocation="netex_nm_singleJourneyPath_support.xsd"/>
 	<xsd:include schemaLocation="netex_nm_singleJourneyService_support.xsd"/>
 	<xsd:include schemaLocation="../../netex_framework/netex_reusableComponents/netex_vehicleType_support.xsd"/>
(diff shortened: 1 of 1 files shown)


skinkie:
Could you elaborate - for a noob like me - why adding an include fixes your issue?

TuThoThai:
> Could you elaborate - for a noob like me - why adding an include fixes your issue?

You're def' not a noob! 😅 

Each XSD file requires that we start by declaring other packages it builds upon. In this specific case, if we call for `VehicleRef`, we need to make sure that `Vehicle` can be found in the dependencies.
Very similarily that in XML files, we check that all `xxxRef` are declared as `xxx` either in the same file or related files.

Hope it helps!

skinkie:
Not yet. Wat is the bug that you have if this is not included? Is this related to XMLSpy or something else?

TuThoThai:
> Not yet. Wat is the bug that you have if this is not included? Is this related to XMLSpy or something else?

It creates 2 types of bugs:
- for softwares that "read" the XSD, such as XMLSpy, it is flag as an unauthorized value in the model
- for some producing a XML file using that part of the XSD, it flags the file as faulty

And, for all who directly generate NeTEx files with code (such as you), it does not change anything. I think.

skinkie:
Does it happen with xmllint? Do you have an example xml file that fails for you prior to this patch?

skinkie (approved):

skinkie:
It was more that my recollection of #921 was that there question was to add VehicleRef (or VehicleTypeRef) and the solution of an xsd include was not what I expected.

PR #937, opened 2025-07-08
A few little bugs

trurlurl:
GaragePointGroup and ParkingPointGroup are never used.

@nick-knowles I changed type of *Covered* in *DeckSpace*.

PR #938, opened 2025-07-08
Updated NeTEx VDV Example

trurlurl:
Complements PR #937.


skinkie:
@trurlurl this won't work, you would need to do this in the other branch.

PR #944, opened 2025-07-28
Extend audiovisual destination information for vehicles

Robbendebiene:
Resolves #882

PR #945, opened 2025-07-28
Remove wrong `PlaceName` property from `HeadingSign`

Robbendebiene:
While working at #944 I noticed that in order to add any `HeadingSign` specific property (like `LineName` or `TransportSubmode`) I had to define a `PlaceName` first. This is probably a copy paste error from `PlaceSign`. At least to me it doesn't make sense why there should be a mandatory `PlaceName` property that must exist in order to define any properties from the `HeadingSignGroup`.

It is probably to late but since this would be a braking change perhaps it could be included in 2.0. Otherwise I could also deprecate the property and set it to ` minOccurs="0"`. Let me know what you think @Aurige @skinkie

### DIFF
--- a/xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentSigns_version.xsd
+++ b/xsd/netex_part_1/part1_ifopt/netex_ifopt_equipmentSigns_version.xsd
@@ -270,11 +270,6 @@ Rail transport, Roads and Road transport
 							<xsd:group ref="SignEquipmentGroup"/>
 						</xsd:sequence>
 						<xsd:sequence>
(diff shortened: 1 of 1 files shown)


Aurige:
To me it should not be mandatory, but I would still keep (that's often that a HEADSIGN is the name of the destination place).

Robbendebiene:
@Aurige Okay thanks for the clarification. I re-introduced the `PlaceName` property, but made it optional. I also moved it to the `HeadingSignGroup` and slightly altered its documentation.

Aurige:
26/08 : need Check with example

skinkie:
(and it is **currently not checked** within github!)

Issue #875, opened 2025-04-22
Allow adding equipment to `VehicleType`


Robbendebiene:
Currently only facilities can be assigned to a `VehicleType` (via `VehicleType/facilities`) but not equipment. Equipment requires a `DeckPlan` element (`EquipableSpace`) or `Vehicle` that references it (via `actualVehicleEquipments`). 

In our opinion any equipment that can be assigned to a `Vehicle` like `AccessVehicleEquipment` or `WheelchairVehicleEquipment`, likewise make sense on a `VehicleType`. Therefore we propose adding the `actualVehicleEquipments` property from `Vehicle` to `VehicleType` in order to allow assigning equipment to a `VehicleType`.

Having `VehicleType/actualVehicleEquipments` containing `EquipmentRef` or `Equipment` also allows defining the equipment directly on the `VehicleType` indicating it is somewhere available on the vehicle. When using `DeckPlan` it can then additionally be referenced or "instantiated" by a particular `EquipableSpace`. Since a vehicle equipment is usually only used at one particular `VehicleType` and not throughout multiple, defining it on the `VehicleType` makes sense and also improves the portability of the `VehicleType` data.
## Questions and remarks
- Would `DeckPlan` elements be able to reference equipment **defined** in a `VehicleType` as assumed above?
- What is the use case of `VehicleTypeRef` in `ActualVehicleEquipmentGroup`? Why can an equipment reference a `VehicleType`?
- There is a small naming inconsistency (same name but different content):
	- `Vehicle/actualVehicleEquipments` contains `EquipmentRef` or `Equipment`
	- `EquipableSpaceGroup/actualVehicleEquipments` contains `ActualVehicleEquipment` 

   To be more consistent `actualVehicleEquipments` from `Vehicle` should be named `vehicleEquipments` following the existing `placeEquipments` property, but I suppose it is too late for this change.

Robbendebiene:
Just want to give a short comparison of the current status quo vs the proposal.
Consider the use case of depicting that a vehicle type has 3 steps in its entrance.

## Status quo

```xml
<ResourceFrame version="any" id="ResourceFrame/id/1">
  <equipments>
    <AccessVehicleEquipment version="any" id="AccessVehicleEquipment/id/1">
      <NumberOfSteps>3</NumberOfSteps>
    </AccessVehicleEquipment>
  </equipments>

  <vehicleTypes>
    <VehicleType version="any" id="VehicleType/id/1">
[...]

PR #947, opened 2025-07-31
Wrong assignment of ActualVehicleEquipmentGroup

Robbendebiene:
I noticed that `EquipmentRef` can be assigned to `WheelchairVehicleEquipment` and `AccessVehicleEquipment`. This seemed wrong to me. My guess is that the `ActualVehicleEquipmentGroup` shouldn't be part of them.

Currently things like this can be done:
```xml
<ActualVehicleEquipment>
   <Units>3</Units>
   <AccessVehicleEquipmentRef ref="a"/>
</ActualVehicleEquipment>

<AccessVehicleEquipment version="any" id="a">
   <Units>3</Units><!-- feels wrong -->
   <BedEquipmentRef ref="x"/><!-- feels wrong -->
   <!-- equipment properties here -->
</AccessVehicleEquipment>
```
I think only this should be possible
```xml
<ActualVehicleEquipment>
   <Units>3</Units>
   <AccessVehicleEquipmentRef ref="a"/>
</ActualVehicleEquipment>

<AccessVehicleEquipment version="any" id="a">
   <!-- equipment properties here -->
</AccessVehicleEquipment>
```

The [commit](https://github.com/NeTEx-CEN/NeTEx/commit/c07c8850e009b308f23e6a3fdecd903d97d6479b) states that these changes are originally from @nick-knowles

### DIFF
--- a/xsd/netex_framework/netex_reusableComponents/netex_equipmentVehiclePassenger_version.xsd
+++ b/xsd/netex_framework/netex_reusableComponents/netex_equipmentVehiclePassenger_version.xsd
@@ -133,6 +133,33 @@ Rail transport, Roads and Road transport
 			<xsd:element ref="AccessibilityAssessment" minOccurs="0"/>
 		</xsd:sequence>
 	</xsd:group>
+	<xsd:group name="_DeprecatedActualVehicleEquipmentGroup">
+		<xsd:annotation>
+			<xsd:documentation>Helper group to retain backwards compatibility with deprecation notice. Will be removed in 3.0.</xsd:documentation>
+		</xsd:annotation>
+		<xsd:sequence>
+				<xsd:element name="Units" type="xsd:integer" minOccurs="0">
+					<xsd:annotation>
+						<xsd:documentation>DEPRECATED: Use Units from ACTUAL VEHICLE EQUIPMENT. Will be removed in 3.0.</xsd:documentation>
+					</xsd:annotation>
+				</xsd:element>
+				<xsd:element ref="VehicleTypeRef" minOccurs="0">
+					<xsd:annotation>
+						<xsd:documentation>DEPRECATED: Use VehicleTypeRef from ACTUAL VEHICLE EQUIPMENT. Will be removed in 3.0.</xsd:documentation>
+					</xsd:annotation>
+				</xsd:element>
+				<xsd:element ref="EquipmentRef" minOccurs="0">
+					<xsd:annotation>
+						<xsd:documentation>DEPRECATED: This should NOT be used. Will be removed in 3.0.</xsd:documentation>
+					</xsd:annotation>
+				</xsd:element>
+				<xsd:element ref="AccessibilityAssessment" minOccurs="0">
+					<xsd:annotation>
+						<xsd:documentation>DEPRECATED: Use AccessibilityAssessment from ACTUAL VEHICLE EQUIPMENT or on the actual element like DECK ENTRANCE. Will be removed in 3.0.</xsd:documentation>
+					</xsd:annotation>
+				</xsd:element>
+		</xsd:sequence>
+	</xsd:group>
 	<!-- ======================================================================= -->
 	<xsd:element name="AccessVehicleEquipment" substitutionGroup="PassengerEquipment">
 		<xsd:annotation>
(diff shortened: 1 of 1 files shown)


ue71603:
@skinkie  you disagree it seems

Aurige:
This is coming from Transmodel ... If I remember well the key point of the ActualVehicleEquipment is to connect the Equipment to a VehicleType, but as the ActualVehicleEquipment can be standalone, it also needs to be able to refer an Equipment (not to be used when already embedded in  an Equipment). But I'm puzzled by the VehicleType relation (instead of Vehicle) in the XSD ...
So changes here would require a larger group discussion
image

Robbendebiene:
I feel like there might be a small misunderstanding, so I just want to restate that the problem I'm reporting is more about `AccessVehicleEquipment` and `WheelchairVehicleEquipment` than about `ActualVehicleEquipment`.

Simply speaking I think `AccessVehicleEquipment` and `WheelchairVehicleEquipment` should not incorporate any properties from `ActualVehicleEquipment` (like the `EquipmentRef`). Instead `ActualVehicleEquipment` should be used to reference `AccessVehicleEquipment` and `WheelchairVehicleEquipment`.

I think `ActualVehicleEquipment` is fine except that it doesn't make sense to allow it in `ResourceFrame/equipments` and that there is a small inconsistency in what `actualVehicleEquipments` can contain which I described here: https://github.com/NeTEx-CEN/NeTEx/issues/875

> But I'm puzzled by the VehicleType relation (instead of Vehicle) in the XSD ...

That is actually a question I had as well which I raised here https://github.com/NeTEx-CEN/NeTEx/issues/875#issue-3010873449 So perhaps it makes sense to discuss these topics together.

Aurige (dismissed):

TuThoThai:
Happy to approve if @skinkie running the script confirms that no examples are broken

skinkie:
THIS FAILS.

examples/functions/timetable/Netex_21_Rail_NetworkTimetable_eurostar.xml:11730: Schemas validity error : Element '{http://www.netex.org.uk/netex}AccessibilityAssessment': This element is not expected. Expected is one of ( {http://www.netex.org.uk/netex}validityConditions, {http://www.netex.org.uk/netex}ValidBetween, {http://www.netex.org.uk/netex}alternativeTexts, {http://www.netex.org.uk/netex}keyList, {http://www.netex.org.uk/netex}privateCodes, {http://www.netex.org.uk/netex}Extensions, {http://www.netex.org.uk/netex}BrandingRef, {http://www.netex.org.uk/netex}Name, {http://www.netex.org.uk/netex}PrivateCode, {http://www.netex.org.uk/netex}PublicCode ).


Where it fails:
```xml
<WheelchairVehicleEquipment version="any" id="est:WheelchairVehicleEquipment:sfe_veheq_trne_LP01_10_02">
        <AccessibilityAssessment version="any" id="est:acaa_sfe_veheq_trne_LP01_10_02">
                <MobilityImpairedAccess>true</MobilityImpairedAccess>
                <suitabilities>
                        <Suitability id="est:acsst_sfe_veheq_trne_LP01_10_02">
                                <MobilityNeed>wheelchair</MobilityNeed>
                                <Suitable>notSuitable</Suitable>
                        </Suitability>
                </suitabilities>
        </AccessibilityAssessment>
        <NumberOfWheelchairAreas>1</NumberOfWheelchairAreas>
        <WidthOfAccessArea>1.0</WidthOfAccessArea>
        <CompanionSeat>true</CompanionSeat>
</WheelchairVehicleEquipment>
```

Robbendebiene:
The online view of Transmodel 2021 (https://transmodel-cen.eu/model/) also contains the inheritance of `ActualVehicleEquipment`. So this at least explains why it is in NeTEx as well.

grafik


The question basically boils down to:
Should `AccessibilityAssessment` be on the equipment definition as well (like it is e.g. for `SanitaryEquipment`) or should it only be on `ActualVehicleEquipment`?
From a `DeckPlan` perspective one always requires an `ActualVehicleEquipment` so it is not needed in this area.
However if the equipment is assigned via an `VehicleEquipmentProfile` or directly to a `Trailing/TractiveElementType` (as in the failing example) referencing an `ActualVehicleEquipment` which in turn references another equipment feels odd.

skinkie:
I am all for deprecating erronous modeling.

Issue #927, opened 2025-06-25
VDV examples


skinkie:
So someone "broke" or at least "was able to spot issues" in the VDV examples.

https://github.com/NeTEx-CEN/NeTEx/actions/runs/15869257615/job/44741970679#step:6:215

Aurige:
it looks that master is broken for 4 month (since https://github.com/NeTEx-CEN/NeTEx/pull/837 ) when no error was triggered ! 
The merge of https://github.com/NeTEx-CEN/NeTEx/pull/925 (comment update) to master, triggered a new validation that caught the error...
Does this makes sense ?

Aurige:
If that's the case, I'm not willing to correct examples in master, so the solution would be to revert https://github.com/NeTEx-CEN/NeTEx/pull/837 and migrate it to next.

Aurige:
Sure I can do it, but wanted first to make sure that this option was OK for the group

Aurige:
@duexw to fix the VDV example with help if @Aurige if needed

duexw:
Fixed in my own clone duexw/NeTEx, but I currently do not have the access right to fix it in this repository. Also I observed that Altova XML Spy 2025 does not find the validation error unless you change the unique constraint to include the namespace of the element, e.g <xsd:field xpath="RegistrationNumber"/> to <xsd:field xpath="netex:RegistrationNumber"/>

PR #953, opened 2025-08-26
Fixed issue 927: Added unique Vehicle/RegistrationNumber

duexw:
This is a fix for issue #927 in master. A unique Vehicle/RegistrationNumber was added where missing


skinkie:
I think this bug must also fix something else: the check that this value should not be empty.

ue71603 (approved):

trurlurl (approved):

skinkie (approved):

TuThoThai:
@skinkie to fix the conflict

TuThoThai:
@duexw, I believe we need to do this one again but targetting either the branch `v2.0` or `v2.0-wip`

PR #954, opened 2025-08-29
Make `TransportType` abstract

Robbendebiene:
`TransportType` is declared abstract in Transmodel. As far as I understand it acts as a base type for `VehicleType`, `Train` etc.

Also something like `ref="TransportType"` does not exist anywhere in the NeTEx schema. So there is also no way to define a `TransportType`.


skinkie:
I think the DummyGroup is the Abstract here.

Robbendebiene:
True but that is because it is a `_DummyType`. At least it doesn't affect the `TransportType` which is why I can do this

```xml
<vehicleTypes>
  <TransportType>
	...  
  </TransportType>
</vehicleTypes>
```

PR #961, opened 2025-10-05
Fix nameofclass nameofrefclass

skinkie:
Still requires a big fight with XML over extension types that cannot be overwritten.


skinkie:
Before @Aurige asks `fixed="Operation"` "Operation" does not exist as class.
Unless someone is clear about abstract, references, etc. the classes are part of the list.

PR #974, opened 2025-10-14
Adding replacement-submodes

JohanEntur:
Due to actual traffic needs, we would like to add a few more enums to NeTEx to properly cover the granulation of transport offers.

This also proves my old point about the need to reform TransportMode/TransportSubmode :)

### DIFF
--- a/xsd/netex_framework/netex_reusableComponents/netex_submode_version.xsd
+++ b/xsd/netex_framework/netex_reusableComponents/netex_submode_version.xsd
@@ -301,6 +301,8 @@ Local train adapted for running in mountain railway lines.</xsd:documentation>
 			<xsd:enumeration value="schoolBus"/>
 			<xsd:enumeration value="schoolAndPublicServiceBus"/>
 			<xsd:enumeration value="railReplacementBus"/>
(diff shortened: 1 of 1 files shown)


TuThoThai (dismissed):
Approved during the plenary of 21 November 2025 but under these conditions:
- NeTEx v2.0 needs to be released first
- PR has to target the future branch v2.1
- This is a quick fix before work v3.0 during with the work on TransportMode will be carrried out

trurlurl (approved):

ue71603 (dismissed):
we once had a railReplacementHelicopter :-)

skinkie:
Why not take the approach where we remove all 'replacements' and have a separate 'ReplacementTransportType' that can link to any known modes?

ue71603:
We could, but if we go down that road, then we would perhaps to think about ALL "actions" that can happen to one or several journeys.

JohanEntur:
This was just an emergency fix imho. In my suggestions regarding TransportMode, this problem would no longer be relevant.

TuThoThai (dismissed):

TuThoThai:
@JohanEntur, I think this PR should be redone targeting either `v2.0` or `v2.1-wip`

PR #970, opened 2025-10-13
Fix required idtypes

skinkie:
The following list is still a todo.
```
NOT Point
NOT GroupMember
NOT GroupConstraintMember
NOT Zone
NOT CommonSectionPointMember
NOT SectionInSequence
NOT SpatialFeature
NOT GeneralFrameMember
NOT Entity_Entity
NOT TypeOfEntity
NOT AddressablePlace
NOT SimpleAvailabilityCondition
NOT ActualVehicleEquipment
NOT TypeOfRollingStock
NOT PassengerAtStopTime
NOT DatedPassingTime
NOT LineSectionPointMember
NOT SiteStructure
NOT LevelInStructure
NOT TicketingEquipment
NOT TicketValidatorEquipment
NOT SignEquipment
NOT AlternativeQuayDescriptor
NOT CustomerService
NOT RechargingPointAssignment
```

Zone and Equipment changes cause failures maybe @ue71603 can help?

```diff
diff --git a/xsd/netex_framework/netex_genericFramework/netex_zone_version.xsd b/xsd/netex_framework/netex_genericFramework/netex_zone_version.xsd
index 97bb08a0..29bd1a98 100644
--- a/xsd/netex_framework/netex_genericFramework/netex_zone_version.xsd
+++ b/xsd/netex_framework/netex_genericFramework/netex_zone_version.xsd
@@ -93,10 +93,34 @@ Rail transport, Roads and Road transport
 		</xsd:complexContent>
 	</xsd:complexType>
 	<!-- ==== ZONE ===================================================== -->
-	<xsd:element name="Zone" type="Zone_VersionStructure" substitutionGroup="GroupOfEntities">
+	<xsd:element name="Zone" substitutionGroup="GroupOfEntities">
 		<xsd:annotation>
 			<xsd:documentation>A two-dimensional PLACE within the service area of a public transport operator (administrative zone, TARIFF ZONE, ACCESS ZONE, etc.).</xsd:documentation>
 		</xsd:annotation>
+		<xsd:complexType>
+			<xsd:complexContent>
+				<xsd:restriction base="Zone_VersionStructure">
+					<xsd:sequence>
+						<xsd:sequence>
+							<xsd:group ref="EntityInVersionGroup" minOccurs="0"/>
+						</xsd:sequence>
+						<xsd:sequence>
+							<xsd:group ref="DataManagedObjectGroup"/>
+						</xsd:sequence>
+						<xsd:sequence>
[...]

PR #977, opened 2025-10-23
Remove required 'id' attribute from accessibility schema

TuThoThai:
Based on the merge of #970, I have looked more into details in elements that I believe do not need an "id" since they are not referened anywhere else.
One element addressed in this PR is "AccessibilityLimitation". As it is embedded within an physical item for which it describes its accessibility, I believe that having an "id" in the first place was a bug in the XSD. It does not appear in the documentation.

### DIFF
--- a/xsd/netex_framework/netex_genericFramework/netex_accessibility/netex_acsb_accessibility.xsd
+++ b/xsd/netex_framework/netex_genericFramework/netex_accessibility/netex_acsb_accessibility.xsd
@@ -90,7 +90,6 @@ Rail transport, Roads and Road transport
 							<xsd:group ref="SensoryLimitationGroup"/>
 						</xsd:sequence>
 					</xsd:sequence>
-					<xsd:attribute name="id" type="LimitationIdType" use="required"/>
 				</xsd:restriction>
 			</xsd:complexContent>
 		</xsd:complexType>


xsd/netex_framework/netex_genericFramework/netex_accessibility/netex_acsb_accessibility.xsd
```diff
@@ -90,7 +90,6 @@ Rail transport, Roads and Road transport
 							<xsd:group ref="SensoryLimitationGroup"/>
 						</xsd:sequence>
 					</xsd:sequence>
-					<xsd:attribute name="id" type="LimitationIdType" use="required"/>
```
skinkie, line 93: We must leave it here. Because a Notice can be assigned via NoticeAssignment, which makes _much_ sense for an AccessibilityLimitation.

TuThoThai: But, then you would have a NoticeAssignment only for "big" objects such as a station, a line, a quay, etc. Not every single object such as an Entrance (for example). So, if we keep an `id`, then it should be at least optional.... 
And, in the case of a NoticeAssignement, I would actually use the id of `AccessibilityAssessment` and not `AccessibilityLimitation`.

ue71603: I don't know where we want to put the Assignment.

TuThoThai: @ue71603 and @skinkie don't you think that if all the files had this, it would be too "noisy"? Because I do not see the point of having both the `id` at the `AccessibilityAssessment ` and `AccessibilityLimitation` levels.
```xml
<Quay id="FR:Quay:n803" version="1">
	<Name>Whatever this Quay is</Name>
	<Centroid>
		<Location id="n803">
			<Longitude>2.65543949937</Longitude>
			<Latitude>48.83601309759</Latitude>
		</Location>
	</Centroid>
	<AccessibilityAssessment id="AccessibilityAssessment:8:" version="any">
		<MobilityImpairedAccess>unknown</MobilityImpairedAccess>
		<limitations>
			<AccessibilityLimitation id="AccessibilityLimitation:8:">
				<WheelchairAccess>true</WheelchairAccess>
				<AudibleSignalsAvailable>false</AudibleSignalsAvailable>
				<VisualSignsAvailable>true</VisualSignsAvailable>
				<TactileGuidanceAvailable>unknown</TactileGuidanceAvailable>
			</AccessibilityLimitation>
		</limitations>
	</AccessibilityAssessment>
	<SiteRef ref=StopPlace:g_n803_bus" version="any"/>
	<TransportMode>bus</TransportMode>
</Quay>
```

skinkie: For a better reason: there doesn't exist an **AccessibilityLimitationRef**.

skinkie (approved):

TuThoThai:
@ue71603, @trurlurl or @thbar: may I get your review on this one?

ue71603 (approved):

TuThoThai:
Merged based on confirmation that id is not needed for AccessibilityLimitation as not referenced elsewhere

PR #988, opened 2026-02-09
Update netex_typeOfValue_version.xsd

### LINKED ISSUE #983: Problems with NeTEx_publication.xsd under XMLSpy
We have a problem with the latest XMLSpy.
It is unhappy about
```
		<xsd:unique name="TypeOfValue_Unique">
			<xsd:selector xpath="values/TypeOfValue/Id"/>
			<xsd:field xpath="."/>
		</xsd:unique>
		<xsd:key name="TypeOfValue_Key">
			<xsd:selector xpath="values/TypeOfValue/Id"/>
			<xsd:field xpath="."/>
		</xsd:key>
	</xsd:element>
```

in
`.xsd\netex_framework\netex_responsibility\netex_typeOfValue_version.xsd`
Image

ue71603:
XMLSpy always had a problem.
This would fix it. But did I do it right? and then I would need to apply it to master too.


thbar:
I'm not well versed enough to understand if that's good or not (and time is limited before release).

I wonder if that's something that could be delayed to `v2.0.1` ?

cc @TuThoThai your opinion on this too?

ue71603:
We certainly can delay it. The new XMLSpy always complaints about the XSD with two warnings.

thbar:
Thank you @ue71603 ; this will give us time for testing. We'll create a milestone and move it to there.

TuThoThai (dismissed):

Aurige:
Why do we have these constraints here ? Shouldn't they be in Publication as all the other ?
The only happens a few time
_**\netex_framework\netex_responsibility\netex_responsibility_version.xsd**_
	(Ln 182, Char 4)     <xsd:unique name="KeyValuePair">
	(Ln 231, Char 4)     <xsd:unique name="PrivateCodeType">

_**\netex_framework\netex_responsibility\netex_typeOfValue_version.xsd**_
	(Ln 178, Char 4)     <xsd:unique name="TypeOfValue_Unique">

_**\netex_part_2\part2_frames\netex_serviceFrame_version.xsd**_
	(Ln 130, Char 4)     <xsd:unique name="SfUniqueLineId">

_**\netex_part_2\part2_frames\netex_serviceFrame_version.xsd**_
	(Ln 130, Char 4)     <xsd:unique name="SfUniqueLineId">

This is not harming, but to be consistent and avoid any double in the future, it may be worth moving them all to NeTEx_Publication.xsd

TuThoThai:
> SfUniqueLineId

Commit https://github.com/NeTEx-CEN/NeTEx/commit/6778f149c67398a421667ecb7a628051a2513a86 addresses this comment

xsd/NeTEx_publication.xsd
```diff
@@ -600,6 +617,16 @@
 			<xsd:field xpath="@id"/>
 			<xsd:field xpath="@version"/>
 		</xsd:key>
+		<!-- =====TypeOfValue unique========================== -->
+		<xsd:unique name="TypeOfValue_Unique">
+			<xsd:selector xpath="values/TypeOfValue"/>
```
Aurige:
```suggestion
			<xsd:selector xpath=".//values/TypeOfValue"/>
```

xsd/NeTEx_publication.xsd
```diff
@@ -600,6 +617,16 @@
...
+			<xsd:field xpath="@id"/>
+		</xsd:unique>
+		<!-- =====TypeOfValue Key========================== -->		
+		<xsd:key name="TypeOfValue_Key">
+			<xsd:selector xpath="values/TypeOfValue"/>
```
Aurige:
```suggestion
			<xsd:selector xpath=".//values/TypeOfValue"/>
```

xsd/NeTEx_publication.xsd
```diff
@@ -600,6 +617,16 @@
 			<xsd:field xpath="@id"/>
 			<xsd:field xpath="@version"/>
 		</xsd:key>
+		<!-- =====TypeOfValue unique========================== -->
+		<xsd:unique name="TypeOfValue_Unique">
```
Aurige: `unique `and `key `on the same attribute is redundant (`key `is enough)

xsd/NeTEx_publication.xsd
skinkie: Are we sure this works at this point? I don't know for sure, but shouldn't this be modeled like KeyValue?

Aurige: not sure what you mean here, KeyValue pairs don't have Id

TuThoThai:
After discussion with @thbar, moving this to v2.0.1 to give us time to examine @skinkie latest comment and ensuring the CI is not failing

thbar:
@TuThoThai I think we have more work on this to do ; given the need to ship, are you ok to move this to `v2.0.1`? EDIT: crossed messages 😄

ue71603:
This is not really a KeyValue standard thing, but only uniqueness within the Type.
I already agreed to do it in 2.0.1.
However, when this is not solved all new XMLspys will report a warning an every time open that file as well and be there. So, to work with it, it is annoying.

ue71603:
We should not have added the other problems to this PR which was on typeOfValue.

However, we are here now, and I think we just have to remove this:
```

		<!-- =====KeyValuePair unique============================== -->
		<xsd:unique name="KeyValuePair">
			<xsd:annotation>
				<xsd:documentation>Every Key Value Pair must be unique.</xsd:documentation>
			</xsd:annotation>
			<xsd:selector xpath=".//netex:KeyValue"/>
			<xsd:field xpath="netex:Key"/>
			<xsd:field xpath="netex:Value"/>
		</xsd:unique>
		<!-- =====PrivateCodeType unique============================== -->
		<xsd:unique name="PrivateCodeType">
			<xsd:annotation>
				<xsd:documentation>Every PrivateCode type must be unique.</xsd:documentation>
			</xsd:annotation>
			<xsd:selector xpath=".//netex:PrivateCode"/>
			<xsd:field xpath="@type"/>
		</xsd:unique>

```
This is in my view semantically wrong for KeyValue which are not primary keys and also probably for PrivateCode.

ue71603:
PrivateCode is even worse. Because most will not use type it will try to do this across the board. For the id-attribute at least the class is used as well.
so I definitely want those to constraints gone.

TuThoThai:
As of 13 March 2026:
- roll back this PR to only address the error netex_typeOfValue_version.xsd
- start a new issue / pull request to address the need (or not) of a constraint on `KeyValue` & `PrivateCodeType` and tag it for v3.0

To be done by @ue71603

xsd/NeTEx_publication.xsd
trurlurl: Similar comment as @Aurige - what is the purpose of unique as we have key? And the non-recursive xpath in unique?

ue71603: fixed

ue71603:
@thbar @TuThoThai  also ready. how does the parent-branch-is-up-to-date work? How can I do this myself.

thbar:
> how does the parent-branch-is-up-to-date work? How can I do this myself.

@ue71603 I figured out what's happening: the check should not run when a PR is targeting `v2.0`, only for `v2.1-wip` or `v3.0-wip` (because the idea is to ensure the WIP branches are always up-to-date from their respective parents, before we merge something in).

It is a matter of changing the ruleset, so that it includes `v2.1-wip` and `v3.0-wip`, but not `v2.0`.

I will fix it on my end then will document things a bit better, so that it's easier for anyone to look into this.

TuThoThai (approved):
Thanks for the work on this one, @ue71603!

PR #995, opened 2026-02-23
Add v2.0.0 to changelog (+ include v1.3.1)

thbar:
Here is a proposal of changelog for `v2.0.0`. 

Most bits are handwritten, some are AI-based: please submit suggestions so that we can have something good enough (cc @TuThoThai @Aurige @Ulf9 @skinkie @ue71603 in particular), thank you!

Meta: once merged, it will be copied to create the GitHub release. This way people cloning the repo have everything, and if we move off GitHub at some point we will already have the data.

Validation side-note: I was about to include a demo of `xmllint` use to support users doing XSD validation, but I think it would drive people to malware stuff (a lot of crap is available in the first Google results) and most importantly, `libxml2` which brings `xmllint` is not maintained anymore since December 2025 (the maintainer has given up due to lack of funding).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,10 +12,10 @@ If you think you found a bug, or have a question, [please provide feedback here]
 
 ### Structural changes:
 
-* BREAKING: many version schemas now enforce stricter validation semantics: mandatory id/ref/version attributes were introduced or tightened in numerous object/ref structures.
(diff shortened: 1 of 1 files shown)


TuThoThai (approved):
LGTM

Aurige (approved):

ue71603 (approved):

PR #1001, opened 2026-03-05
Fix-Buggy SIRI import in occupancy

Aurige:
The SIRI import is probably coming from a SIRI copy/paste since the NeTEx occupancy is following the SIRI one. 

The _NaturalLanguageStringStructure_ is not NeTEx type, and is only used in this file, and doesn't follow NeTEx style/rules. It's replaced with the more "traditional" _MultilingualString_

The SIRI import is crashing some generic NeTEx import  (especially if SIRI is also used) in some implementation (one of them being OpRa). The reason for this crash is difficult to analyse (and is partially depending on the validation engine), but is basically related to object redefinition. As a result this import was a useless and dangerous breaking change and has to be removed.

The SIRI import was there only to allow the use of _NaturalLanguageStringStructure_ and can just be removed.

This is a fix 100% backward compatible at XML Level


ue71603 (approved):

ue71603:
wow @Aurige using language reserved to  @skinkie ;-)

TuThoThai (approved):

thbar (approved):
LGTM! I didn't see other similar references.

PR #1008, opened 2026-03-16
new DRT exemple with small zones with quays

prhod:
As discussed in the last week's Flexible Task Force , here is my exemple using the guidelines provided.
Overview, modelization and questions are all stated in the top part of the file.


ue71603 (approved):
I would merge it into the main PR and continue there.

ue71603:
@prhod  thanks for providing the example

PR #1017, opened 2026-05-06
DirectionType allowed in LineInDirectionRef structure

ue71603:
In JourneyPattern DirectionType is allowed instead of Direction. This was not done in LineInDirection. We believe that this harmonisation is needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -88,3 +88,4 @@ fabric.properties
 # Editor-based Rest Client
 .idea/httpRequests
 
+/.idea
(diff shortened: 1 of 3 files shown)


skinkie:
It contains refs, why do you want to have this container to be an enumeration?

ue71603:
We will remove Direction from our profile. We only have two Direction inbound and  outbound (and many others have this too). Using Direction to remodel what DirectionType does is not really useful. And as it was mentioned. In JourneyPattern this exact behaviour is also modeled that way.

skinkie:
And the meaning is now: this interchange applies to all ServiceJourneys with Line+DirectionType or ServiceJourneys with ServiceJourneyPattern+LineRef+DirectionType?

ue71603:
I think different combinations are reasonable:
- LineRef
- AllLines
- LineRef + DirectionRef
- LineRef + DirectionType
- ServiceJourney

TuThoThai:
@ue71603, should not this one target branch `v2.1-wip`?

ue71603:
replaced by https://github.com/TransmodelEcosystem/NeTEx/pull/1021...  

@skinkie I wasn't able to rebase to v2.1-wip from v2.0. Perhaps another issue with the repo?

PR #1019, opened 2026-05-06
Adding some more elements to PassingTime

ue71603:
When switching from Call to PassingTime we found that we might need some things that don't exist there.
We suggest to add those to PassingTime.

However, we have the problem, that the things are on  CallPartStructure as e.e. checkConstraints might be different. We could have an ArrivalCheckConstraint and a DepartureCheckConstraint. What do you think?


skinkie:
Why not in the ServiceJourneyPattern?

ue71603:
wasn't a good fit in my view. Where should I put it in? I would need to put it in PointInJourneyPattern

TuThoThai:
@ue71603, should not this one target `v2.1-wip`?

skinkie:
> wasn't a good fit in my view. Where should I put it in? I would need to put it in PointInJourneyPattern

StopPointInJourneyPattern, then it would be applied to all Journeys = less data to be shipped.

ue71603:
continued here: https://github.com/TransmodelEcosystem/NeTEx/pull/1020

PR #1020, opened 2026-05-06
adding CheckConstraint to StopPointInJourneyPattern and Passingtime

ue71603:
replaces https://github.com/TransmodelEcosystem/NeTEx/pull/1019
We allow the CheckConstraint in two additional places:
* StopPointInJourneyPattern when the CheckConstraint applies to the whole ServiceJourneyPattern
* if however only on a special journey in a greater set of journeys following a ServiceJourneyPattern a special CheckConstraint appears, then it is more effective to put it into Passingtime and leave the JourneyPattern intact.


xsd/netex_part_1/part1_tacticalPlanning/netex_servicePattern_version.xsd
```diff
@@ -1,13 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
+<!-- edited with XMLSpy v2026 (x64) (https://www.altova.com) by Urs Stöckli (Schweizerische Bundesbahnen SBB) -->
```
trurlurl:
```suggestion
```

trurlurl (dismissed):

trurlurl (dismissed):

xsd/netex_part_1/part1_tacticalPlanning/netex_servicePattern_version.xsd
```diff
@@ -901,6 +902,9 @@ Rail transport, Roads and Road transport
 					<xsd:documentation>Nature of use of stop, e.g. access, interchange only, or pass through. Default is Access.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
+			<xsd:group ref="QuayAssignmentGroup"/>
```
skinkie, line 906: Did you place it in the middle to align it with Call?

ue71603: I tried. But different place would also be ok. We had in OJP some cases where we switched the order of some elements in service and this caused problems. So to do it as harmonised as possible seems a good idea to me.

skinkie: You understand that the way you are now doing it explicitly would break parsing and validation? So if it was deliberate to put them here, was this because in the other structures they are already at this point?

ue71603: What does it break? everything is optional in in.

duexw: I do not see that the change would be incompatible in the sense that a pre-existing file would not validate any more. Maybe an existing importer could run into problems if it is not able to ignore unknown elements, e.g. it expects bookingElements to follow directly after AccessibilityAssessment. But this means we can add new optional elements only as the last element of the parent element.

duexw (changes requested):
I am basically OK with the proposal. If we want to change from using Call to using PassingTimes, we need to put CheckConstraint somewhere. However, it would be nicer to have CheckConstraint in PassingTime as well. Why? Because Journeys with the same ServiceJourneyPattern may have different CheckConstraints. E.g. a congestion check constraint may only apply to peak time journeys. With the existing proposal, we would have to create extra ServiceJourneyPatterns in this case.

ue71603:
@duexw  now possible in both places with a caveat on Passingtime

ue71603:
@skinkie @duexw more happy now?

trurlurl (dismissed):

ue71603:
@duexw  do we need multiple CheckConstraint on ServiceJourneyPattern? on ServiceJourney there is a container.

ue71603:
@duexw Is this sufficient for you?

xsd/netex_part_1/part1_tacticalPlanning/netex_servicePattern_version.xsd
```diff
@@ -63,8 +65,7 @@
 					<Requires>http://www.netex.org.uk/schemas/1.0/PATH/netex_prereqfile.xsd</Requires>
 				</Relation>
 				<Rights>Unclassified
-						<Copyright>CEN, Crown Copyright 2009-2023</Copyright>
-				</Rights>
+						<Copyright>CEN, Crown Copyright 2009-2023</Copyright></Rights>
```
skinkie:
```suggestion
						<Copyright>CEN, Crown Copyright 2009-2023</Copyright>
				</Rights>
```

xsd/netex_part_1/part1_tacticalPlanning/netex_servicePattern_version.xsd
```diff
@@ -901,6 +903,13 @@ Rail transport, Roads and Road transport
 					<xsd:documentation>Nature of use of stop, e.g. access, interchange only, or pass through. Default is Access.</xsd:documentation>
 				</xsd:annotation>
 			</xsd:element>
+			<xsd:group ref="QuayAssignmentGroup"/>
+			<xsd:element ref="AccessibilityAssessment" minOccurs="0"/>
+			<xsd:element name="checkConstraints" type="checkConstraints_RelStructure" minOccurs="0">
+				<xsd:annotation>
+					<xsd:documentation>Impediments to navigation from processes or barriers. For example security, check in etc., +v2.1</xsd:documentation>
```
skinkie:
```suggestion
					<xsd:documentation>Implements the navigation from processes or barriers. For example security, check in etc., +v2.1</xsd:documentation>
```

duexw (dismissed):

TuThoThai (approved):
LGTM after tiny documentation improvements.

I will let @ue71603, @skinkie and @duexw double-check.

skinkie (approved):

ue71603:
so, ready to merge...

Issue #1018, opened 2026-05-06
CI/build issue after moving org


thbar:
Reported by @ue71603 

<img width="1682" height="746" alt="Image" src="https://github.com/user-attachments/assets/03f367a4-bf39-43cb-9a6e-be803d99bab2" />

thbar:
Things I noticed so far:
- the failing PR (https://github.com/TransmodelEcosystem/NeTEx/pull/1017) has been created 2 hours ago, so _after_ the Github org was migrated (https://github.com/TransmodelEcosystem/NeTEx/issues/1011)
- the test done this Sunday for the migration was done on an existing PR (https://github.com/TransmodelEcosystem/NeTEx/pull/998/changes/af348027bc11d9b7e45cbe49480ccd6a21a23884), and was successful

Will continue investigating.

TuThoThai:
Also to note, on the same day of the move, I launched a PR for which the CI worked. However, the PR was only in documentation part of a single XSD. PR was #1015

TuThoThai:
@thbar, I found it. This is the wrong setting of the repo but I do not know how to address it alone as it seems I cannot choose the first option.

<img width="843" height="366" alt="Image" src="https://github.com/user-attachments/assets/dbb1643c-bc40-4294-b0ca-6ef7d7d764ee" />

thbar:
@TuThoThai calling you ;-)

thbar:
Closer zoom - it's the commit step, happening only when we have changes if I understand well:

<img width="1286" height="406" alt="Image" src="https://github.com/user-attachments/assets/38adf60c-f97f-4d71-bea1-74ee1119e228" />

thbar:
https://github.com/TransmodelEcosystem/NeTEx/commit/dc3fbf0c27cf984470ed3b8ef26cf6528f8dfeb8 fixed the reproduced failure (or at least a first part of it).

TuThoThai:
Updates done to branches `v2.1-wip` and `v3.0-wip` for them to be on-par with branch `v2.0` including commit https://github.com/TransmodelEcosystem/NeTEx/commit/dc3fbf0c27cf984470ed3b8ef26cf6528f8dfeb8

CI has worked on both PR #1020 and #1021

Issue #982, opened 2026-01-13
[EPIC] Proposal for the backport of missing commits from `master` into `next`


thbar:
Currently, 32 commits have been added to `master`, which are not yet into `next` (or not with the same commit ids, at least). My understanding is that we need to bring back those changes into `next` before we can [release NeTEx v2.0](https://github.com/NeTEx-CEN/NeTEx/issues/963).

<img width="522" height="191" alt="Image" src="https://github.com/user-attachments/assets/b5aef1a3-d5e6-4e1e-bc61-0d5a530d2ce1" />

I’m here first listing those commits to create a todo list, then proposing a way to gradually backport those changes safely and in reviewed manner, into `next`.
### Commits from master not yet into next, by chronological order

I generated the list so we can track our work here, using this (+ a bit of manual editing):

``` 
git log master ^next --pretty=format:"%h %ad %an|%s" --date=iso-strict --reverse | sed -E 's/^(.*)\|.* \(#([0-9]+)\)$/\1 (#\2)/'
```

This gives our nice 32-commits todo list so we can follow our progress:

- [x] d847bb7 2024-03-25T17:41:09+01:00 Stefan de Konink (#690)
- [x] 4bf30d8 2024-04-15T20:20:32+02:00 Stefan de Konink (#686)
- [x] a3ba1e5 2024-04-17T17:23:30+02:00 Matthias Günter (#612)
- [x] de9d705 2024-04-17T17:29:22+02:00 Stefan de Konink (#464)
- [x] 56046b2 2024-04-17T17:33:43+02:00 Stefan de Konink (#454)
- [x] 7ce0095 2024-04-17T23:50:42+02:00 Matthias Günter (#640)
- [x] af04aa1 2024-05-01T17:33:47+02:00 Stefan de Konink (#462)
- [x] 84c52d1 2024-05-02T13:36:17+02:00 Stefan de Konink (#705)
- [x] 64aaec7 2024-05-02T13:37:04+02:00 Julius Tens (#696)
- [x] 6eb10a8 2024-05-03T15:58:53+02:00 Stefan de Konink (#722)
- [x] 91b24bd 2024-05-03T18:29:24+02:00 Matthias Günter (#720)
- [x] 0c59d9e 2024-05-03T18:30:32+02:00 Matthias Günter (#724)
- [x] bcdfecb 2024-05-03T18:31:02+02:00 Stefan de Konink (#723)
[...]

PR #1022, opened 2026-05-07
Make sure PR cannot be merged, if their base branch is not up-to-date itself (avoids #982)

thbar:
If it works properly, it will ultimately warn us before merging a PR, and avoid troubles like on #1020 (which is not up-to-date from v2.0, which has the fix for #1018).

The goal is to avoid backport hell (#982), & make sure we sync on a regular basis (how we will do it is another problem, I can do it manually for now @TuThoThai).

So the idea is:
- merge this
- then make sure v2.1-wip is made up-to-date from v2.0 (this will carry the #1018 fix)
- then #1020 should become green!

CC @TuThoThai


TuThoThai (approved):
Excellent work!

ue71603 (approved):

PR #1031, opened 2026-06-17
Add StopPointInServiceJourneyPatternRef to JourneyWaitTime for stops visited 2 times in the same ServiceJourney with different wait times

ue71603:
In some cases the WaitTime is different if a stop is visited multiple times within a ServiceJourney. When modeled with TimeDemandType then two WaitTimes are needed e.g. first visit 1 minute and second visit 4 minutes, because change of driver.

In those cases (and only those) a StopPointInServiceJourneyPatternRef should be filled in.

Is this ok? Other ideas?


Joostb61:
Do you really have 2 WaitingTimes at the same stop in one journey? Can you give an example from Switzerland? In NL we have 2 or more TimeDemandTypes, but this is not on the same journey.

ue71603:
Wilfried's example:
example_2_wait_times
 

The journeys visit "Avenches, Gare" twice, obviously the vehicle does a loop. For whatever reason they do not have a waiting time at the first visit, but at the second visit they wait for 4 minutes. This situation occurs quite often in regional bus services. How do you express this using time demand type? In VDV 462 it works, because we put the waitTimes Element below StopPointInJourneyPattern.

hilcokats:
In the Netherlands we solved this exact case by requiring two disctinct ScheduledStopPoints A and B linked to the same Quay in their respective PassengerStopAssignments. In that case, you can link a WaitTime X to SSP A and a WaitTime Y to SSP B.

ue71603:
@duexw  Whatdoyou think? I don't think it should be done that way. All other attributes are not in ScheduledStopPoint that are relevant for single journeys.

ue71603:
I had a discussion with @skinkie and he thinks we should it do as proposed  here and not with two different ScheduledStopPoints.

duexw:
We have the same issue in VDV 452, and up to now it is often solved by creating a "virtual" stop point for the second visit. However, this introduces follow up problems, as in reality there is only one stop point. For instance, in AVM systems you have to assign a different stop id to the virtual stop point, this again does not match anything found in the central stop register and so on. I therefore approve of the solution proposed here.

duexw (dismissed):

xsd/netex_part_1/part1_tacticalPlanning/netex_journeyTiming_version.xsd
```diff
@@ -158,6 +159,11 @@ Rail transport, Roads and Road transport
 		</xsd:annotation>
 		<xsd:sequence>
 			<xsd:element ref="TimingPointRef" minOccurs="0"/>
+			<xsd:element ref="StopPointInJourneyPatternRef" minOccurs="0">
```
skinkie: I don't think this should be a sequence. I think it should be a choice. I think I can also make an argument for TimingPointInJourneyPatternRef.

ue71603: @skinkie  TimingPointRef is always present. This is in our examples a ScheduledStopPoint (as that's the level where the TimingLinks are described mostly. Borderpoint is also possible.  So this allows us for the WaitTime to anchored to something that remains mostly static.

For JourneyPattern variations we need an additional element. I modeled it now as a choice.

Indirectly one could go to the ScheduledStopPoint also thought those structures. However, if somebody has not implemented this edge case, then they are lost. So I prefer to have to have the TimingPointRef separatly.

Can you live with that? PR updated.

xsd/netex_part_1/part1_tacticalPlanning/netex_journeyTiming_version.xsd
```diff
@@ -158,6 +159,14 @@ Rail transport, Roads and Road transport
 		</xsd:annotation>
 		<xsd:sequence>
 			<xsd:element ref="TimingPointRef" minOccurs="0"/>
+			<xsd:choice minOccurs="0">
+				<xsd:annotation>
+					<xsd:documentation>In some cases there are multiple visits to the same stop with a different WaitTime (e.g. on the second there is a change in the driver. Many systems don't support this feature in providing the data. So this element should be used with care. +v2.1</xsd:documentation>
+				</xsd:annotation>
+				<xsd:element ref="PointInJourneyPatternRef"/>
```
skinkie: I wonder... can you do this just PointInJourneyPatternRef?

skinkie (changes requested):
I would want a small example here too.

duexw:
Hi Stefan, 
I agree choice would be better than sequence. I made a small example when I did the change to StopPointInJourneyPatternRef (see also attachment): 
image
[NeTEx_CH_Bern_Spiez_Zweisimmen_ValidRef.xml](https://github.com/user-attachments/files/29752912/NeTEx_CH_Bern_Spiez_Zweisimmen_ValidRef.xml)

However, I would not allow for wait times at any type of point, as it would open quite a few issues.

xsd/netex_part_1/part1_tacticalPlanning/netex_journeyTiming_version.xsd
```diff
@@ -158,6 +159,14 @@ Rail transport, Roads and Road transport
 		</xsd:annotation>
 		<xsd:sequence>
 			<xsd:element ref="TimingPointRef" minOccurs="0"/>
+			<xsd:choice minOccurs="0">
```
skinkie, line 161: @duexw @ue71603 This should still change.

Make sure the TimingPointRef is part of the choice, and while we are at it, I think adding `ScheduledStopPointRef` would make sense for me (that would be more a NeTEx 3.0 addition, but I hope you understand my I mention it now).

My persional preference would also be that the minOccurs=0 is then removed, but that is likely not possible do the JourneyWaitTimeGroup is also included some other thing that already species the Point.

ue71603: I  can't add ScheduledStopPointRef as this is a TimingPointRef as well.

ue71603: I added it to the choice as Wilfried wants it too. So it is always either.

xsd/netex_part_1/part1_tacticalPlanning/netex_journeyTiming_version.xsd
```diff
@@ -157,7 +158,26 @@ Rail transport, Roads and Road transport
 			<xsd:documentation>Elements of a JOURNEY WAIT TIME.</xsd:documentation>
 		</xsd:annotation>
 		<xsd:sequence>
-			<xsd:element ref="TimingPointRef" minOccurs="0"/>
+			<xsd:choice minOccurs="0">
+				<xsd:annotation>
+					<xsd:documentation>In some cases there are multiple visits to the same stop with a different WaitTime (e.g. on the second there is a change in the driver. Then one of the PointInJourneyPatternRef is needed instead of the TimingPoint. &gt;Must be optional, because in the JourneyPatternWaittime other element make the reference to the pattern. Otherwise a TimingPointRef is the way to use.  +v2.1</xsd:documentation>
```
skinkie:
```suggestion
					<xsd:documentation>In some cases there are multiple visits to the same stop with a different WaitTime (e.g. on the second there is a change in the driver. Then one of the PointInJourneyPatternRef is needed instead of the TimingPoint. Must be optional, because in the JourneyPatternWaittime other element make the reference to the pattern. Otherwise a TimingPointRef is the way to use. +v2.1</xsd:documentation>
```

xsd/netex_part_1/part1_tacticalPlanning/netex_journeyTiming_version.xsd
```diff
@@ -157,7 +158,26 @@ Rail transport, Roads and Road transport
 			<xsd:documentation>Elements of a JOURNEY WAIT TIME.</xsd:documentation>
 		</xsd:annotation>
 		<xsd:sequence>
-			<xsd:element ref="TimingPointRef" minOccurs="0"/>
+			<xsd:choice minOccurs="0">
+				<xsd:annotation>
+					<xsd:documentation>In some cases there are multiple visits to the same stop with a different WaitTime (e.g. on the second there is a change in the driver. Then one of the PointInJourneyPatternRef is needed instead of the TimingPoint. Must be optional, because in the JourneyPatternWaittime other element make the reference to the pattern. Otherwise a TimingPointRef is the way to use. +v2.1</xsd:documentation>
```
skinkie:
```suggestion
					<xsd:documentation>In some cases there are multiple visits to the same stop with a different WaitTime (e.g. on the second there is a change in the driver). Then one of the PointInJourneyPatternRef is needed instead of the TimingPoint. Must be optional, because in the JourneyPatternWaittime other element make the reference to the pattern. Otherwise a TimingPointRef is the way to use. +v2.1</xsd:documentation>
```

skinkie (dismissed):

duexw (dismissed):

TuThoThai (approved):
LGTM after a tiny documentation correction

TuThoThai:
This PR was previously approved by @duexw and @skinkie
I merge it after I have added my own review that includes a documentation change + addressing the previously failed GitHub Action

skinkie:
@ue71603 @TuThoThai @duexw for documentation perspective there is an implication with addition. This means that a TimeDemandType gets closely coupled with one (or more) ServiceJourneyPatterns, while there hasn't been an explicit relation before. The only relation existed because a ServiceJourney would specify a ServiceJourneyPattern and a TimeDemandType. Technically the relationship between ServiceJourneyPattern and TimeDemandType does not have to exist. Nothing would stop you to model the entire timing model for all links in the network in one TimeDemandType. That this in practise never happens has to do with the fact that transit modellers don't publish true timing but a percentille with respect to the start of the trip.

ue71603:
@skinkie I agree that these only should be used when it is absolutely necessary (e.g. when there are two visits). Do I need to make the documentation comment clearer in that respect?

skinkie:
I don't really care that there is an alternative way to model it. What we need to clear about is how the relation between ServiceJourneyPattern and TimeDemandType _should_ work. That is a bigger topic than this pull request.

skinkie:
@ue71603 the point being ;-) you are at an earlier commit than I am now. I updated to v2.0.

ue71603:
I understand. I will use the latest as well and commit the new one to the branch.

PR #998, opened 2026-02-26
Remove unique constraints, because we have xsd:key

skinkie:
Fix #994

With unique constraints:
```
real	4m5.527s
user	3m57.850s
sys	0m1.706s
```

**Without** unique constraints:
```
real	2m44.075s
user	2m39.062s
sys	0m1.093s
```

PR #721, opened 2024-05-02
Merge master to next 20240502

skinkie:
This is a merge review for master to next. I want have it properly reviewed since there are some rough edges. Please remove anything unrelated to the merge.

PR #1032, opened 2026-06-20
Restore 4 uniqueness removed in #998 where no xsd:key covers it (+ must-fail guards)

thbar:
**Scope of this PR**

- To be merged into #998 ("Remove unique constraints, because we have xsd:key").
- Undoes 4 of #998's removals — the cases where no `xsd:key` enforces the uniqueness: `GroupOfLinkSequences`, `OperatingDay`, `ValidBetween` and `ValidityPeriod`.
- Adds must-fail examples (`examples/should-fail/`) that prove the uniqueness is no longer enforced — we can decide later what to do with them (e.g. keep them as permanent CI guards).

This PR **restores the missing `xsd:unique`** — otherwise `v2.0.1` would not be a *correct* optimisation, since it would silently drop validation that `v2.0.0` had (these four were enforced in `v2.0.0` via `xsd:unique`).

It does **not** replace them with `xsd:key`. For `GroupOfLinkSequences`, afaik `xsd:key` was there before `v2.0.0`,  but has removed by #721/#462 before `v2.0.0` was released (which is why it's important to avoid also removing unique here).

Whether to (re)introduce keys is a separate discussion we can have, but not the scope of this PR (and maybe not for a `v2.0.1` bug fix either).

### DIFF
--- a/xsd/NeTEx_publication.xsd
+++ b/xsd/NeTEx_publication.xsd
@@ -2257,6 +2257,15 @@
 			<xsd:field xpath="@version"/>
 		</xsd:key>
 		<!-- =====ValidityCondition============================== -->
+		<!-- =====ValidityCondition unique========================== -->
+		<xsd:unique name="ValidityCondition_UniqueBy_Id_Version">
+			<xsd:annotation>
+				<xsd:documentation>Every [ValidityCondition Id + Version] must be unique within document.</xsd:documentation>
+			</xsd:annotation>
+			<xsd:selector xpath=".//netex:ValidityCondition | .//netex:ValidityTrigger | .//netex:ValidityRuleParameter | .//netex:ValidBetween"/>
+			<xsd:field xpath="@id"/>
+			<xsd:field xpath="@version"/>
+		</xsd:unique>
 		<!-- =====ValidityCondition Key ========================== -->
 		<xsd:keyref name="ValidityCondition_KeyRef" refer="netex:ValidityCondition_AnyVersionedKey">
 			<xsd:selector xpath=".//netex:ValidityConditionRef | .//netex:ValidityTriggerRef | .//netex:ValidityRuleParameterRef | .//netex:WithConditionRef"/>
--- a/xsd/NeTEx_publication.xsd
+++ b/xsd/NeTEx_publication.xsd
@@ -979,6 +979,15 @@
 			<xsd:field xpath="@version"/>
 		</xsd:key>
 		<!-- =====GroupOfLinkSequences============================== -->
+		<!-- =====GroupOfLinkSequences unique========================== -->
+		<xsd:unique name="GroupOfLinkSequences_UniqueBy_Id_Version">
+			<xsd:annotation>
+				<xsd:documentation>Every [GroupOfLinkSequences Id + Version] must be unique within document.</xsd:documentation>
+			</xsd:annotation>
+			<xsd:selector xpath=".//netex:GroupOfLinkSequences"/>
+			<xsd:field xpath="@id"/>
+			<xsd:field xpath="@version"/>
+		</xsd:unique>
 		<!-- =====SimpleFeature============================== -->
 		<!-- =====SimpleFeature Key ========================== -->
 		<xsd:keyref name="SimpleFeature_KeyRef" refer="netex:SimpleFeature_AnyVersionedKey">
--- a/xsd/NeTEx_publication.xsd
+++ b/xsd/NeTEx_publication.xsd
@@ -6515,6 +6515,15 @@
 			<xsd:field xpath="@version"/>
 		</xsd:key>
 		<!-- ===== UsageValidityPeriod Key ========================== -->
+		<!-- ===== UsageValidityPeriodting unique========================== -->
+		<xsd:unique name="UsageValidityPeriod">
+			<xsd:annotation>
+				<xsd:documentation>Every [UsageValidityPeriod Id + Version] must be unique within document.</xsd:documentation>
+			</xsd:annotation>
+			<xsd:selector xpath=".//netex:UsageValidityPeriod | .//netex:ValidityPeriod"/>
+			<xsd:field xpath="@id"/>
+			<xsd:field xpath="@version"/>
+		</xsd:unique>
 		<xsd:keyref name="UsageValidityPeriod_KeyRef" refer="netex:UsageValidityPeriod_AnyVersionedKey">
 			<xsd:selector xpath=".//netex:UsageValidityPeriodRef"/>
 			<xsd:field xpath="@ref"/>
--- a/xsd/NeTEx_publication.xsd
+++ b/xsd/NeTEx_publication.xsd
@@ -2200,6 +2200,15 @@
 			<xsd:field xpath="@version"/>
 		</xsd:key>
 		<!-- =====OperatingDay============================== -->
+		<!-- =====OperatingDay unique==In Calendar======================== -->
+		<xsd:unique name="OperatingDay_UniqueCalendarDateInCalendar">
+			<xsd:annotation>
+				<xsd:documentation>Each date is only allowed once per calendar.</xsd:documentation>
+			</xsd:annotation>
+			<xsd:selector xpath=".//netex:ServiceCalendarFrame/netex:operatingDays/netex:OperatingDay"/>
+			<xsd:field xpath="netex:CalendarDate"/>
+			<xsd:field xpath="@version"/>
+		</xsd:unique>
 		<!-- =====OperatingDay Key ========================== -->
 		<xsd:keyref name="OperatingDay_KeyRef" refer="netex:OperatingDay_AnyVersionedKey">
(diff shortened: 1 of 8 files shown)


xsd/NeTEx_publication.xsd
```diff
@@ -979,6 +979,15 @@
 			<xsd:field xpath="@version"/>
 		</xsd:key>
 		<!-- =====GroupOfLinkSequences============================== -->
+		<!-- =====GroupOfLinkSequences unique========================== -->
+		<xsd:unique name="GroupOfLinkSequences_UniqueBy_Id_Version">
```
skinkie, line 983: Why do you think this is correct? For example, why isn't this just a key?

xsd/NeTEx_publication.xsd
```diff
@@ -6515,6 +6515,15 @@
 			<xsd:field xpath="@version"/>
 		</xsd:key>
 		<!-- ===== UsageValidityPeriod Key ========================== -->
+		<!-- ===== UsageValidityPeriodting unique========================== -->
+		<xsd:unique name="UsageValidityPeriod">
```
skinkie, line 6519: It should be unique by key.

skinkie:
@thbar read the comments.

xsd/NeTEx_publication.xsd
thbar: See PR description (just added) ; I think it's more correct that what is in #998 on that specific bit, which was effectively removal of any sort of uniqueness check. 

Why not key, well, key has been removed before `v2.0.0` (see description). I don't know if adding it back in a patch release is a good idea or not, we'll have to discuss that collectively.

xsd/NeTEx_publication.xsd
thbar: See PR description, I'm just undoing problematic removal of any uniqueness at all, this is not a PR where I'm changing any unique to a key (and as commented above, at least one was removed just pre-`v2.0.0`, so it would be a bit confusing).

To be discussed, then (but this is not this PR's goal).

thbar:
> @thbar read the comments.

I got notified by your other comments, no problem. Now the PR is ready to review.

Hopefully we can merge & we didn't miss any other similar issues, so that #998 can go ahead.

xsd/NeTEx_publication.xsd
skinkie: A duplicate id is never correct, it is a key. Hence it should not be unique.

xsd/NeTEx_publication.xsd
```diff
@@ -2239,6 +2257,15 @@
 			<xsd:field xpath="@version"/>
 		</xsd:key>
 		<!-- =====ValidityCondition============================== -->
+		<!-- =====ValidityCondition unique========================== -->
+		<xsd:unique name="ValidityCondition_UniqueBy_Id_Version">
```
skinkie, line 2261: Should just be a key.

thbar: This is worth a dedicated ticket (for this one & all the places deserving a key, either removed erroneously before `v2.0.0` or not), and its own separate PR(s).

#1032 is just short rollback of 4 missing uniques.

xsd/NeTEx_publication.xsd
thbar: See https://github.com/TransmodelEcosystem/NeTEx/pull/1032#discussion_r3453549474

xsd/NeTEx_publication.xsd
skinkie: I think the opposite is true. It was erronousley added as unique, and the removal script removed id unique, because the expectation was a key must exist for an id.

thbar: Yes, yet the key isn't there, so it creates a larger trouble even (no more unicity at all). We'll need to address that, so I created https://github.com/TransmodelEcosystem/NeTEx/issues/1034 ; we'll catalogue over there.

xsd/NeTEx_publication.xsd
TuThoThai: @thbar and @skinkie, 
I confirm that in v1.3, this part was written as follow:
```xml
		<!-- =====GroupOfLinkSequences============================== -->
		<!-- =====GroupOfLinkSequences unique========================== -->
		<xsd:unique name="GroupOfLinkSequences_UniqueBy_Id_Version">
			<xsd:annotation>
				<xsd:documentation>Every [GroupOfLinkSequences Id + Version] must be unique within document.</xsd:documentation>
			</xsd:annotation>
			<xsd:selector xpath=".//netex:GroupOfLinkSequences"/>
			<xsd:field xpath="@id"/>
			<xsd:field xpath="@version"/>
		</xsd:unique>
```
Direct link is [this one](https://github.com/TransmodelEcosystem/NeTEx/blob/4f42794047ec9944fd38e4497f3af142a33462c8/xsd/NeTEx_publication.xsd#L1363).

Considering that `GroupOfLinkSequencesRef` is based on the substitution group `VersionOfObjectRef` and not `TypeOfEntityRef`, I am uncertain that a key will work.

However, let's make it simple and roll back to the original state of this part of the XSD.

xsd/NeTEx_publication.xsd
TuThoThai: @thbar and @skinkie,
`ValidityCondition` has both the `unique` and `key` checked. So we can omit this roll-back.

XSD from v1.3 is as follow
```xml
		<!-- =====ValidityCondition unique========================== -->
		<xsd:unique name="ValidityCondition_UniqueBy_Id_Version">
			<xsd:annotation>
				<xsd:documentation>Every [ValidityCondition Id + Version] must be unique within document.</xsd:documentation>
			</xsd:annotation>
			<xsd:selector xpath=".//netex:ValidityCondition | .//netex:ValidityTrigger | .//netex:ValidityRuleParameter | .//netex:ValidBetween"/>
			<xsd:field xpath="@id"/>
			<xsd:field xpath="@version"/>
		</xsd:unique>
		<!-- =====ValidityCondition Key ========================== -->
		<xsd:keyref name="ValidityCondition_KeyRef" refer="netex:ValidityCondition_AnyVersionedKey">
			<xsd:selector xpath=".//netex:ValidityConditionRef | .//netex:ValidityTriggerRef | .//netex:ValidityRuleParameterRef | .//netex:WithConditionRef"/>
			<xsd:field xpath="@ref"/>
			<xsd:field xpath="@version"/>
		</xsd:keyref>
		<xsd:key name="ValidityCondition_AnyVersionedKey">
			<xsd:selector xpath=".//netex:ValidityCondition | .//netex:ValidityTrigger | .//netex:ValidityRuleParameter | .//netex:AvailabilityCondition"/>
			<xsd:field xpath="@id"/>
			<xsd:field xpath="@version"/>
		</xsd:key>

```

This is the [permalink](https://github.com/TransmodelEcosystem/NeTEx/blob/4f42794047ec9944fd38e4497f3af142a33462c8/xsd/NeTEx_publication.xsd#L3462).

xsd/NeTEx_publication.xsd
TuThoThai: Indeed, `UsageValidityPeriod` has both unique and `key` checked. So, the stand-alone `key` can remain.

XSD v1.3:
```xml
		<!-- ===== UsageValidityPeriod Key ========================== -->
		<!-- ===== UsageValidityPeriodting unique========================== -->
		<xsd:unique name="UsageValidityPeriod">
			<xsd:annotation>
				<xsd:documentation>Every [UsageValidityPeriod Id + Version] must be unique within document.</xsd:documentation>
			</xsd:annotation>
			<xsd:selector xpath=".//netex:UsageValidityPeriod | .//netex:ValidityPeriod"/>
			<xsd:field xpath="@id"/>
			<xsd:field xpath="@version"/>
		</xsd:unique>
		<xsd:keyref name="UsageValidityPeriod_KeyRef" refer="netex:UsageValidityPeriod_AnyVersionedKey">
			<xsd:selector xpath=".//netex:UsageValidityPeriodRef"/>
			<xsd:field xpath="@ref"/>
			<xsd:field xpath="@version"/>
		</xsd:keyref>
		<xsd:key name="UsageValidityPeriod_AnyVersionedKey">
			<xsd:selector xpath=".//netex:UsageValidityPeriod"/>
			<xsd:field xpath="@id"/>
			<xsd:field xpath="@version"/>
		</xsd:key>
```

[Permalink](https://github.com/TransmodelEcosystem/NeTEx/blob/4f42794047ec9944fd38e4497f3af142a33462c8/xsd/NeTEx_publication.xsd#L9746) to the XSD

TuThoThai (changes requested):
@thbar, I believe that the only roll-back needed is on GroupOfLinkSequences since we have key checked for the other 3.

Could you please double-check using your examples? 🙏

thbar:
After double checking with @Aurige & @TuThoThai, 3 of the 4 "failing examples" are correct (which means that we are indeed removing too many uniques & they should be rolled back).

But this PR does too much now : reverting 4 uniques, + adding a framework of test & failing examples. As advised by @Aurige I'm closing this one, in favour of a PR introducing only the 3 "should fail" examples and the test harness for now:
- #1074

PR #1054, opened 2026-07-29
Removed excessive spaces from enum values

### LINKED ISSUE #1053: Some enumeration values in examples contain extra spaces
Some files contain enumeration values with extra spaces.

While testing my parser I have found 91 occurences. For example:

https://github.com/TransmodelEcosystem/NeTEx/blob/cc8dbf9e9476be1caaf01a44cdb754de2405e88e/examples/functions/deckPlans/DeckPlans-Example_train.xml#L4642

Should be just `left`, without the trailing space.

I will make a fixing PR later today.

Micovec:
While testing my parser for NeTEx I have found these enum values that contain extra spaces. Because of that the parser cannot parse the files.

Fixes: #1053

### DIFF
--- a/examples/functions/deckPlans/DeckPlans-Example_train.xml
+++ b/examples/functions/deckPlans/DeckPlans-Example_train.xml
@@ -4639,7 +4639,7 @@ The train reverse in  the station and departs, oriented backwards to the right
 								</PathLinkInSequence>
 								<PathLinkInSequence version="any" id="xx:xx:PLIS-station_765@SPE-entrance_1+BP-quay_2_boarding_point_c@by_stairs-2">
 									<SitePathLinkRef version="any" ref="xx:SPL-station_765@PJ-ticket_hall_middle+SPE-stairs_to_q1_and_q2_top"/>
-									<Heading>left </Heading>
+									<Heading>left</Heading>
 									<Transition>level</Transition>
 									<Instruction>Turn left and proceed to entrance to platforms 1 and  2</Instruction>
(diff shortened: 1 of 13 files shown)


skinkie:
@Micovec while they are not pretty, they are not _wrong_ right? I think your parser should handle this.

Micovec:
@skinkie I guess you're right. I did not look up the type of these enumerations. They are xsd:NMTOKEN which allows leading and triling spaces.

I'm closing this PR and will try to resolve it in the parser.

skinkie:
@thbar @TuThoThai  I think we should fix this, but leave open an example to tests parsers.

Micovec:
The reason I closed this PR was due to the fact that C# class for deserializing xml files does not support xsd:NMTOKEN for enumerations.

I could circumvent it in the parser if the owner of [XmlSchemaClassGenerator](https://github.com/mganss/XmlSchemaClassGenerator) would approve my suggestion (I have not yet created an issue in that repo). I use that project to generate files for parsing.

Another fix would be to ask .NET developers for a support of xsd:NMTOKEN but I doubt that will go anywhere. It's an unusual request. Besides the class for deserializing definitely does not support the whole xsd.

If you are willing to fix it I would be more than happy.

skinkie:
> If you are willing to fix it I would be more than happy.

The point is, your parser needs to be able to parse valid XML. Otherwise it will be again a specialistic attempt having too many limitations. I'll try to send you an email.

PR #1056, opened 2026-08-05
the attribute in the relationshipStructure should not be id  but name…

ue71603:
in reality only name of classes are allowed and so the attribut can't be called id, but should be called nameOfClass.

as in most cases like <delays> nobody uses id, but if you do, then it must be a NameOfClass.

### DIFF
--- a/xsd/netex_framework/netex_responsibility/netex_relationship.xsd
+++ b/xsd/netex_framework/netex_responsibility/netex_relationship.xsd
@@ -146,9 +146,9 @@ Rail transport, Roads and Road transport
 		</xsd:annotation>
 	</xsd:element>
 	<!-- =====Metamodel references================================================ -->
-	<xsd:simpleType name="RelationshipId">
+	<xsd:simpleType name="RelationshipNameOfClass">
 		<xsd:annotation>
-			<xsd:documentation>Type for Identifier of a relationship.</xsd:documentation>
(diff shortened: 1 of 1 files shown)


xsd/netex_framework/netex_responsibility/netex_relationship.xsd
```diff
@@ -195,7 +195,7 @@ Rail transport, Roads and Road transport
 		<xsd:annotation>
 			<xsd:documentation>Abstract Type for a serialisation of a NeTEx relationship.</xsd:documentation>
 		</xsd:annotation>
-		<xsd:attribute name="id" type="RelationshipId">
+		<xsd:attribute name="nameOfClass" type="RelationshipId">
```
skinkie: Don't think this is right.

ue71603: I think it is. have a look what RelationshipId really is.

ue71603: it is derived from NameOfClass.

skinkie: RelationshipId isn't the right name... but you are right then.

ue71603:
I try to change the name and we see if something breaks.

TuThoThai (approved):

PR #1066, opened 2026-08-20
Fix misleading ParticipantRef documentation

hfjelstad:
## Problem

The `ParticipantRef` element in the publication schemas is documented as
*"Identifier of system requesting Data."*

That wording only fits the request case. In a publication delivery the field
identifies the party that *produced* the data, not a requestor — so the
mandatory (delivery) use is documented incorrectly.

Additionally, the value is a **code** (NMTOKEN via `siri:ParticipantCodeType`),
not a reference to an object. There is no `Participant` type and no `keyref` —
the value *is* the identity. The `...Ref` name suggests an object reference that
does not exist.

## Fix

Short term (documentation only, non-breaking): update the `xsd:documentation`
to cover both publication and request:

> External code identifying the source of the data being exchanged.

Applied to all four top-level publication schemas:
- `xsd/NeTEx_publication.xsd`
- `xsd/NeTEx_publication_timetable.xsd`
- `xsd/NeTEx_publication-NoConstraint.xsd`
- `xsd/wsdl/NeTEx_publication-NoConstraint.xsd`

The element name and type are left unchanged, so there is no schema break.

### DIFF
--- a/xsd/NeTEx_publication-NoConstraint.xsd
+++ b/xsd/NeTEx_publication-NoConstraint.xsd
@@ -66,7 +66,7 @@ Roads and Road transport.
 	<!-- ======================================================================= -->
 	<xsd:element name="ParticipantRef" type="siri:ParticipantCodeType">
 		<xsd:annotation>
-			<xsd:documentation>Identifier of system requesting Data.</xsd:documentation>
+			<xsd:documentation>External code identifying the source of the data being exchanged.</xsd:documentation>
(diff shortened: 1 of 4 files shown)


skinkie (dismissed):
Maybe do a pull request on the SIRI repo too?

trurlurl (approved):
To me, the following would make things even more clear: "External code identifying the sender of the message."

TuThoThai (approved):

PR #1074, opened 2026-09-17
"Should fail" examples framework

thbar:
As suggested by @Aurige, I'm moving out the new "should fail test harness" (introduced in #1032, related to #998) in a separate PR for clarity.

This PR introduces negative examples: examples that should not be considered XSD-valid.

(only 3 to get our feet wet & to separate concerns with #998/#1032). 

Assertions checks have been narrowed compared to original implementation.


TuThoThai (approved):
Many thanks for this work, @thbar! And, we should definitely explore further how to add other types of must-fail examples. 

Note: this change ill have to be backported to branches `v2.1-wip`, `v2.1-wip-efip`, and `v3.0-wip`

ue71603 (approved):

Aurige (approved):

